PR #1, opened 2020-09-03
Implement PIDForest using PyCall

janfrancu:
Feel free to comment the code in the changed files and not just the changes themselves, as this pull request does not contain all of them needed for pidforest.


vitskvara:
I won't comment on the content of the `pidforest` model, just keep the fit script up to date with the API, e.g. https://github.com/aicenter/GenerativeAD.jl/blob/73617dede83b55da32583cfa1a5fa6eac8d9bc80/scripts/experiments_tabular/knn.jl#L103 needs an additional argument, the edit_params function

janfrancu:
Ready for review.
Note that this does not include the setup of the virtualenv and installation of pidforest, as it should be added in some setup script.
Also the environment should be relative to the current repository and not fixed as it is now to my home.

src/models/pidforest.jl
```diff
@@ -9,17 +9,21 @@ mutable struct PIDForest
...
 function StatsBase.fit!(model::PIDForest, X::Array{T, 2}) where T<:Real
+	# no need for data transposition as the forest is learned on column major
     model.forest.fit(X)
 end
 
-function StatsBase.predict(model::PIDForest, x; err=0.1, pct=50) where T<:Real
-    -model.forest.predict(x; err=0.1, pct=50)[end] # last element contains scores
+function StatsBase.predict(model::PIDForest, x; pct=50) where T<:Real
+	# anomaly scores correspond to percentile `pct` over trees
+	# the lower the score the more anomalous the sample should be
```
vitskvara, line 27: usually, this is the other way around, i.e. the higher the score, the more anomalous the sample is (think reconstruction error)

janfrancu: I described the internals of pidforest here, where the lower the score the more anomalous the sample is. That's why I've put here the minus sign and that is also what the authors had done when computing the evaluation metrics.
https://github.com/janfrancu/pidforest/blob/master/scripts/run_classification_sklearn_pidforest_alg.py#L75-L82

vitskvara: Ok, that makes sense, I was just trying to make sure we use the same convention

scripts/experiments_tabular/run_parallel.sh
```diff
@@ -1,5 +1,10 @@
...
-MODELSCRIPT=$1
-MAX_SEED=$2
-cat datasets_tabular.txt | xargs -n 1 -P 40 ./${MODELSCRIPT}.sh $MAX_SEED
\ No newline at end of file
+MODEL=$1
+NUM_SAMPLES=$2
+MAX_SEED=$3
+
+while read d; do
+    sbatch --array=1-$NUM_SAMPLES -D /home/francja5/Projects/modules/GenerativeAD/scripts/experiments_tabular/ ./${MODEL}_run.sh $MAX_SEED $d
```
vitskvara: Do the 100 different experiments happen here? Does that mean they are trained in parallel? Then the check_params function is not gonna work, since it can only tell that a model was already trained if it was already saved.

janfrancu: This should schedule an array job (repeating the same command with the same arguments) for each dataset. [AFAIK](https://slurm.schedmd.com/job_array.html) those task in the array are run in parallel, however we can limit the number of concurrent jobs to 1 as in `--array=1-${NUM_SAMPLES}%1`.  In some cases the search space is large and collisions may happen rarely (assuming 100 samples), however I see your point for example with `kNN`, which has really only one parameter. 
What about adding the number of concurrent as a parameter to the script?

vitskvara: That sounds reasonable, the repetition of parameters is probably only gonna happen for simple models where training time is probably not going to be critical. So introducing a new script parameter sounds good.

janfrancu: Added.

PR #14, opened 2020-09-11
New run script

vitskvara:
This includes the following changes:

- ensuring that the same parameters are used across the cross-validation folds
- fixing the check params functions that was broken in case there were duplicates
- adding an argument to the run_parallel.sh script - the file with a list of dataset names

### DIFF
--- a/scripts/experiments_tabular/knn.jl
+++ b/scripts/experiments_tabular/knn.jl
@@ -8,7 +8,7 @@ using BSON
 
 s = ArgParseSettings()
 @add_arg_table! s begin
-   "seed"
+   "max_seed"
         required = true
         arg_type = Int
         help = "seed"
@@ -18,7 +18,7 @@ s = ArgParseSettings()
         help = "dataset"
 end
 parsed_args = parse_args(ARGS, s)
-@unpack dataset, seed = parsed_args
+@unpack dataset, max_seed = parsed_args
 
 #######################################################################################
 ################ THIS PART IS TO BE PROVIDED FOR EACH MODEL SEPARATELY ################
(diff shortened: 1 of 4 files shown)


scripts/experiments_tabular/knn_run.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 #SBATCH --time=24:00:00
-#SBATCH --nodes=1 --ntasks-per-node=1 --cpus-per-task=1
+#SBATCH --nodes=1 --ntasks-per-node=2 --cpus-per-task=1
```
janfrancu, line 3: Have you checked if it helps with the number of submission we can make to the slurm scheduler/queue?

scripts/experiments_tabular/knn.jl
```diff
@@ -80,33 +80,36 @@ end
...
-	if GenerativeAD.check_params(GenerativeAD.edit_params, savepath, data, parameters)
-		# fit
-		training_info, results = fit(data, parameters)
-		# here define what additional info should be saved together with parameters, scores, labels and predict times
-		save_entries = merge(training_info, (modelname = modelname, seed = seed, dataset = dataset))
-		
-		# now loop over all anomaly score funs
-		for result in results
-			GenerativeAD.experiment(result..., data, savepath; save_entries...)
+max_tries = 10*max_seed
```
janfrancu: Is it really necessary to multiply the number of tries? What we are trying here is to just sample 10 times some HP, just to see if one of them runs. I believe nothing has changed about that by introducing the for cycle over seeds.

scripts/experiments_tabular/knn.jl
```diff
@@ -80,33 +80,36 @@ end
...
+			# fit
+			training_info, results = fit(data, parameters)
+			# here define what additional info should be saved together with parameters, scores, labels and predict times
+			save_entries = merge(training_info, (modelname = modelname, seed = seed, dataset = dataset))
+
+			# now loop over all anomaly score funs
+			for result in results
+				GenerativeAD.experiment(result..., data, savepath; save_entries...)
+			end
+			global try_counter = max_tries + 1
```
janfrancu, line 107: `try_counter = max_tries` should be enough to terminate the while cycle

src/experiments.jl
```diff
@@ -51,7 +51,9 @@ function check_params(edit_params_f, savepath, data, parameters)
 	if ~isdir(savepath)
 		return true
 	end
-	saved_params = map(x -> DrWatson.parse_savename(x)[2], readdir(savepath))
+	# filter out duplicates created by tagsave
+	fs = filter(x->!(occursin("_#", x)), readdir(savepath))
```
janfrancu, line 55: What exactly was the problem here? Are there some temporary files that DrWatson creates containing `"_#"`?

janfrancu (commented):
Looks good, just need some clarification here and there.

scripts/experiments_tabular/knn.jl
vitskvara: Yeah thats not necessary, it is a remnant of some testing

scripts/experiments_tabular/knn.jl
vitskvara: Yes, but that triggers the printout on the last line, which should not happen if everything went alright

src/experiments.jl
vitskvara: Yes, if you use tagsave with safesave feature and the new saved file is named the same as an existing file, it appends "_#1" or some other number. But then parse_savename cannot parse this file and everything crashes :-)

scripts/experiments_tabular/knn_run.sh
vitskvara: Not yet, I think this should just speed things a little bit.

scripts/experiments_tabular/knn.jl
janfrancu: Oh I see.

scripts/experiments_tabular/knn.jl
janfrancu: Please resolve it soon, so that it can be merged.

janfrancu (approved):

PR #17, opened 2020-09-11
Add edit_params for pidforest.

### LINKED ISSUE #15: Overload edit_params with PIDforest
Edit the HP number of samples for tree building according to the size of the dataset.

janfrancu:
Also modified the sampling ranges for hyper parameters.

### DIFF
--- a/scripts/experiments_tabular/pidforest.jl
+++ b/scripts/experiments_tabular/pidforest.jl
@@ -1,13 +1,13 @@
-using ArgParse
-using GenerativeAD
-import StatsBase: fit!, predict, sample
 using DrWatson
 @quickactivate
+using ArgParse
+using GenerativeAD
+using StatsBase: fit!, predict, sample
 using BSON
 
 s = ArgParseSettings()
 @add_arg_table! s begin
-   "seed"
+   "max_seed"
         required = true
         arg_type = Int
         help = "seed"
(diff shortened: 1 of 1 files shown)


scripts/experiments_tabular/pidforest.jl
```diff
@@ -17,16 +17,27 @@ s = ArgParseSettings()
...
 
 	return (;zip(argnames, map(x->sample(x, 1)[1], par_vec))...)
 end
 
+function GenerativeAD.edit_params(data, parameters)
+	D, N = size(data[1][1])
+	if N < parameters.max_samples
+		# if there are not enough samples, choose closest multiple of 50
+		@info "Not enough samples in training, changing max_samples for each tree."
+		return merge(parameters, (;max_samples = max(25, N - mod(N, 50))))
```
vitskvara, line 35: Is the max_samples parameter going to have a discontinuity at N? Since for N >= max_samples it stays the same, but for N = max_samples - 1 the modulo is subtracted. Might that be a problem?

janfrancu: My thinking was that:
1. if the dataset has less than 50 samples -> choose for each tree 25 samples
2. if the dataset has more than 50 samples but less than supplied max_samples -> choose some closest multiples of 50

The latter should take care of datasets ranging from 50 to hundreds. For bigger datasets there are usually many valid choices of this hyperparameter, so that editting won't take place that often.

PR #18, opened 2020-09-11
Matěj z

TheMrGhostman:
GANomaly now can run on GPU (successfully tested).

Only thing that was not fully tested is appling results into `GenerativeAD.experiment(result..., data, savepath; save_entries...) `


scripts/experiments_images/GANomaly.jl
```diff
@@ -22,10 +23,28 @@ end
...
+            [1:5...],
+            [0.0001:0.0001:0.001..., 0.002:0.001:0.01...],
+            [20],
+            [2^x for x=2:8],
+            )
+	return NamedTuple{argnames}(map(x->sample(x,1)[1], options))
+end
+
+
+function check_params(savepath, parameters, data)
```
vitskvara: you can use GenerativeAD.check_params

scripts/experiments_images/GANomaly.jl
```diff
@@ -42,42 +61,71 @@ Note:
...
-    residue = isize % 16
-    if residue != 0
-        isize = isize + 16 - residue
-        X_train = GenerativeAD.resize_images(X_train, isize, in_ch)
-        X_val = GenerativeAD.resize_images(X_val, isize, in_ch)
-        X_test = GenerativeAD.resize_images(X_test, isize, in_ch)
-    end
     # prepare batches & loaders
-    train_loader = MLDataPattern.RandomBatches(X_train, parameters.batch_size, parameters.iters)
+    train_loader = Flux.Data.DataLoader(data |> gpu, batchsize=parameters.batch_size, shuffle=true)
```
vitskvara: Does moving all the data to gpu at once work? In my experience this leads to using up too much of the card memory (or running out of memory completely). I usually keep the data on cpu memory and only send the current batch to gpu.

vitskvara: You could handle that on the level of your custom fit! function.

scripts/experiments_images/GANomaly.jl
```diff
@@ -42,42 +61,71 @@ Note:
...
+savepath = datadir("experiments/images/$(modelname)/$(dataset)/seed=$(seed)")
 
+data = GenerativeAD.load_data(dataset, seed=seed)
+
+try_counter = 0
+max_tries = 2
+while try_counter < max_tries
+	parameters = sample_params()
+
+    # computing additional parameters
```
vitskvara: All this should be done by overloading the GenerativAD.edit_params function

scripts/experiments_images/GANomaly.jl
```diff
@@ -42,42 +61,71 @@ Note:
...
+    isize = isize + 16 - isize % 16
+    # update parameter
+    parameters = merge(parameters, (isize=isize, in_ch = in_ch, out_ch = 1))
+	# here, check if a model with the same parameters was already tested
+	if check_params(savepath, parameters, data)
+
+        data = GenerativeAD.preprocess_images(data, parameters)
+        #(X_train,_), (X_val, y_val), (X_test, y_test) = data
+
+		training_info, results = fit(data[1][1], parameters)
```
vitskvara: Please fix the indentation to the same format.

src/models/GANomaly.jl
```diff
@@ -266,34 +293,64 @@ function StatsBase.fit!(generator::Generator, discriminator::Discriminator, opt,
...
+            end
+            grad = back(1f0)
+            Flux.Optimise.update!(opt, ps_d, grad)
+
+            history = update_history(history, loss1, loss2)
+            next!(progress; showvalues=[(:epoch, "$(epoch)/$(epochs)"),
+                                        (:generator_loss, loss1[1]),
+                                        (:discriminator_loss, loss2)
+                                        ])
+            #případně přidat kontrolu velikosti chyby s restartem discriminatoru
```
vitskvara: Do you not use early stopping?

scripts/experiments_images/GANomaly.jl
```diff
@@ -1,5 +1,6 @@
 using ArgParse
 using GenerativeAD
+using GenerativeAD.Models: anomaly_score
 using DrWatson
```
vitskvara: Move `using DrWatson` and `@quickactivate` to the first 2 lines of the script, that way it will always trigger the version of code that you run from,

vitskvara: which is useful for testing on RCI cluster in your home folder.

scripts/experiments_images/GANomaly.jl
```diff
@@ -42,42 +61,71 @@ Note:
...
+
+#_________________________________________________________________________________________________
+
+savepath = datadir("experiments/images/$(modelname)/$(dataset)/seed=$(seed)")
 
+data = GenerativeAD.load_data(dataset, seed=seed)
+
+try_counter = 0
+max_tries = 2
+while try_counter < max_tries
```
vitskvara, line 92: Check `experiments_tabular/knn.jl`, we have changed this loop so that the loop over seeds is inside. However, for image datasets, I think it will be more important to do the loop over classes properly and only run over one seed - otherwise its going to be too computationally heavy.

scripts/experiments_images/GANomaly.jl
```diff
@@ -42,42 +61,71 @@ Note:
...
-    return training_info, predict(X_val)
+    return training_info, [(x -> GenerativeAD.Models.anomaly_score(generator|>cpu, x; dims=3), parameters)]
+    # not sure if I should return generator and disciriminator in GPU
+end
+
+#_________________________________________________________________________________________________
+
+savepath = datadir("experiments/images/$(modelname)/$(dataset)/seed=$(seed)")
 
+data = GenerativeAD.load_data(dataset, seed=seed)
```
vitskvara: Loop over classes here, i.e. use `GenerativeAD.load_data(dataset, seed=seed, anomaly_class_ind=i)`

scripts/experiments_images/GANomaly.jl
```diff
@@ -27,96 +31,87 @@ modelname = "Conv-GANomaly"
...
-    return training_info, [(x -> GenerativeAD.Models.anomaly_score(generator|>cpu, x; dims=3), parameters)]
-    # not sure if I should return generator and disciriminator in GPU
+	return training_info, [(x -> GenerativeAD.Models.anomaly_score(generator|>cpu, x; dims=3), parameters)]
+	# not sure if I should return generator and disciriminator in GPU
 end
 
 #_________________________________________________________________________________________________
 
-savepath = datadir("experiments/images/$(modelname)/$(dataset)/seed=$(seed)")
+savepath = datadir("experiments/images/$(modelname)_ac=$(anomaly_class_ind)/$(dataset)/seed=$(seed)")
```
vitskvara: Append the class index to `$dataset`, not to the model name.

vitskvara:
Looks good

scripts/experiments_images/GANomaly.jl
```diff
@@ -10,74 +11,118 @@ using MLDataPattern
...
+function sample_params()
+	argnames = (:latent_dim, :num_filters, :extra_layers, :lr, :batch_size,
+                :iters, :check_every, :patience, )
+	options = (
+			[10:10:200...],
+			[2^x for x=2:8],
+			[0:3...],
+			[0.0001:0.0001:0.001..., 0.002:0.001:0.01...],
+            [2^x for x=2:8],
+			[10000],
```
vitskvara, line 42: No more epochs?

scripts/experiments_images/GANomaly.jl
```diff
@@ -10,74 +11,118 @@ using MLDataPattern
...
-        :history => info[1]
+	training_info = (
+		fit_t = fit_t,
+		model = (generator|>cpu, discriminator|>cpu),
+		history = info[1] # losses through time
 		)
 
-    return training_info, predict(X_val)
+	return training_info, [(x -> GenerativeAD.Models.anomaly_score(generator|>cpu, x; dims=3), parameters)]
+	# not sure if I should return generator and disciriminator in GPU
```
vitskvara: This depends on whether you continue to compute the scores on gpu (which should be faster), but returning the model on cpu is safer, it can always be ported back.

scripts/experiments_images/GANomaly.jl
```diff
@@ -10,74 +11,118 @@ using MLDataPattern
...
+end
+
+#_________________________________________________________________________________________________
+
+
+try_counter = 0
+max_tries = 10*max_seed
+
+
+while try_counter < max_tries
```
vitskvara, line 92: This is just a detail, but the indentation format is still a bit off.

scripts/experiments_images/GANomaly.jl
```diff
@@ -10,74 +11,118 @@ using MLDataPattern
...
+        		data = GenerativeAD.Models.preprocess_images(data, parameters)
+        		#(X_train,_), (X_val, y_val), (X_test, y_test) = data
+                training_info, results = fit(data, parameters)
+
+        		save_entries = merge(training_info, (modelname = modelname, seed = seed, dataset = dataset, anomaly_class = i))
+
+        		# now loop over all anomaly score funs
+        		for result in results
+        			GenerativeAD.experiment(result..., data, savepath; save_entries...)
+        		end
```
vitskvara: You are missing the condition for ending the while loop, see `knn.jl`

scripts/experiments_images/GANomaly_run.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+#SBATCH --time=24:00:00
+#SBATCH --nodes=1 --ntasks-per-node=2 --cpus-per-task=1
+#SBATCH --gres=gpu:1
+#SBATCH --mem=20G
```
vitskvara: Is this enough memory?

scripts/experiments_images/GANomaly_run.sh
```diff
@@ -0,0 +1,14 @@
...
+
+MAX_SEED=$1
+DATASET=$2
+ANOMALY_CLASSES=$3
+
+module load Julia/1.4.1-linux-x86_64
+module load Python/3.8.2-GCCcore-9.3.0
```
vitskvara: Do you need Python? Are you using PyCall?

scripts/experiments_images/SkipGANomaly.jl
```diff
@@ -0,0 +1,127 @@
...
+	try
+		global info, fit_t, _, _, _ = @timed fit!(model |>gpu , data, parameters)
+	catch e
+		println("Error caught.")
+		return (fit_t = NaN,), []
+	end
+
+	training_info = (
+		fit_t = fit_t,
+		model = (model |> cpu),
```
vitskvara: Save model only once. Probably in the fit function in case of a successful fit. That means that the fit function also needs to know the savepath. The filename should consist of the parameters used in the construction of the model - it should not contain the parameters of the anomaly scoring functions. Test whether this design does not cause trouble somewhere else, e.g. in the check params function.

scripts/experiments_images/SkipGANomaly.jl
```diff
@@ -0,0 +1,127 @@
...
+        		data = GenerativeAD.Models.preprocess_images(data, parameters, denominator=32)
+        		#(X_train,_), (X_val, y_val), (X_test, y_test) = data
+                training_info, results = fit(data, parameters)
+
+        		save_entries = merge(training_info, (modelname = modelname, seed = seed, dataset = dataset, anomaly_class = i))
+
+        		# now loop over all anomaly score funs
+        		for result in results
+        			GenerativeAD.experiment(result..., data, savepath; save_entries...)
+        		end
```
vitskvara: Again, missing break of the while loop.

scripts/experiments_images/SkipGANomaly_run.sh
```diff
@@ -0,0 +1,14 @@
...
+#SBATCH --nodes=1 --ntasks-per-node=2 --cpus-per-task=1
+#SBATCH --gres=gpu:1
+#SBATCH --mem=20G
+
+MAX_SEED=$1
+DATASET=$2
+ANOMALY_CLASSES=$3
+
+module load Julia/1.4.1-linux-x86_64
+module load Python/3.8.2-GCCcore-9.3.0
```
vitskvara: Same points as in the case of the ganomaly script

src/models/Models.jl
```diff
@@ -10,8 +10,11 @@ using ImageTransformations
 using MLDataPattern
 using ProgressMeter: Progress, next!
 
+include("utils/utils.jl")
```
vitskvara, line 13: Does this need to be in a subdirectory? I think it only makes sense if we add more similar files there.

src/models/SkipGANomaly.jl
```diff
@@ -0,0 +1,315 @@
...
+		grad = back(1f0)
+		Flux.Optimise.update!(opt, ps_d, grad)
+
+		history = update_history(history, loss1, loss2)
+		next!(progress; showvalues=[(:iters, "$(iters)/$(params.iters)"),
+									(:generator_loss, loss1[1]),
+									(:discriminator_loss, loss2)
+									])
+		#TODO optionaly add discriminator restrart if its loss drops under 1e-5
+		if mod(iters, params.check_every) == 0
```
vitskvara: Inconsistent indentation, a bit hard to read. How does the early stopping work here?

vitskvara (changes requested):

scripts/experiments_images/GANomaly.jl
TheMrGhostman: yes. one epoch was too much for GANomaly (it hits minimum val_loss within 300 iterations but epoch is about 600) . So i had to go back to iterations.

scripts/experiments_images/GANomaly.jl
TheMrGhostman: I thought so. I wanted to avoid unnecessary complications.

vitskvara: Actually when I look at it again it cannot be ported back to gpu when its a part of the anonymous function... maybe leaving it on gpu is better, but you have to take care of having the data on gpu in evaluation as well.

scripts/experiments_images/GANomaly.jl
TheMrGhostman: I know that sometimes I have two empty lines instead of one. 
However i am not sure why this happens 
`
+        		data = GenerativeAD.Models.preprocess_images(data, parameters, denominator=32)
+        		#(X_train,_), (X_val, y_val), (X_test, y_test) = data
+                training_info, results = fit(data, parameters)
`
because in my editor it is correct.

scripts/experiments_images/GANomaly.jl
TheMrGhostman: I missed that ... i added `global try_counter = max_tries + 1`

scripts/experiments_images/GANomaly_run.sh
TheMrGhostman: I think it is, but i can raise it.

scripts/experiments_images/GANomaly_run.sh
TheMrGhostman: good point. I forgot remove it.

scripts/experiments_images/SkipGANomaly.jl
TheMrGhostman: fixed

scripts/experiments_images/SkipGANomaly_run.sh
TheMrGhostman: fixed

src/models/Models.jl
TheMrGhostman: It is not needed but i did it because Honza made this directory in his branch/pull request. So i thought i can just use it.

src/models/SkipGANomaly.jl
TheMrGhostman: Here it is bit tricky(resp messy) but while i checking checking early stopping criterion `anomality = (params.lambda*tot_val_loss_rl + (1-params.lambda)*tot_val_loss_ll)` 
i am also computing generator and discriminator loss on validation dataset with normal observations.
I thought it can be useful to trace it as well.

src/models/Models.jl
vitskvara: yeah then lets keep it

src/models/SkipGANomaly.jl
vitskvara: I see, this makes sense.

vitskvara (approved):

vitskvara (approved):

src/models/utils/utils.jl
```diff
@@ -0,0 +1,92 @@
...
+		isize = isize + denominator - residue
+		X_train = scale_to_interval(resize_images(X_train, isize, in_ch), range=range)
+		X_val = scale_to_interval(resize_images(X_val, isize, in_ch), range=range)
+		X_test = scale_to_interval(resize_images(X_test, isize, in_ch), range=range)
+	end
+
+	return (X_train, y_train), (X_val, y_val), (X_test, y_test)
+end
+
+function GANomalyHistory()
```
vitskvara, line 41: Consider using ValueHistories.jl for this, they basically do exactly the same thing.

scripts/experiments_images/GANomaly.jl
vitskvara: So des that mean you never go through the whole dataset? Is that not a little limiting? Maybe for non-MNIST problems you will have to do epochs.

TheMrGhostman: that is true. Reconstruction capabilities of this autoencoder with those multiple loss functions are pretty good. For other datasets we can simply set more possible iterations. But author of original paper stated that  he used 10, 15 and 25 epochs for non-MNIST datasets (25 epochs = +/- 15 000 iterations)

src/models/utils/utils.jl
TheMrGhostman: i will take a look on it

scripts/experiments_images/GANomaly.jl
```diff
@@ -10,74 +11,118 @@ using MLDataPattern
...
+
+try_counter = 0
+max_tries = 10*max_seed
+
+while try_counter < max_tries
+	parameters = sample_params()
+
+	for seed in 1:max_seed
+		for i in 1:anomaly_classes
+			savepath = datadir("experiments/images/$(modelname)/$(dataset)_ac=$(i)/seed=$(seed)")
```
vitskvara: Change to `"experiments/images/$(modelname)/$(dataset)/ac=$(i)/seed=$(seed)"`

TheMrGhostman: changed

scripts/experiments_images/SkipGANomaly.jl
```diff
@@ -0,0 +1,129 @@
...
+	catch e
+		println("Error caught.")
+		return (fit_t = NaN,), []
+	end
+
+	training_info = (
+		fit_t = fit_t,
+		model = (info[2] |> cpu),
+		history = info[1], # losses through time
+        n_parameters = info[3] # number of parameters
```
vitskvara: please rename this to npars

scripts/experiments_images/SkipGANomaly.jl
```diff
@@ -0,0 +1,129 @@
...
+"""
+function fit(data, parameters)
+	# define models (Generator, Discriminator)
+	model, _ = GenerativeAD.Models.SkipGANomaly_constructor(parameters)
+
+	try
+		global info, fit_t, _, _, _ = @timed fit!(model |>gpu , data, parameters)
+	catch e
+		println("Error caught.")
+		return (fit_t = NaN,), []
```
vitskvara: set the rest of the fields as well, e.g. `model = nothing, ...`

scripts/experiments_images/SkipGANomaly.jl
```diff
@@ -0,0 +1,129 @@
...
+			# update parameter
+			parameters = merge(parameters, (isize=isize, in_ch = in_ch, out_ch = 1))
+			# here, check if a model with the same parameters was already tested
+			if GenerativeAD.check_params(savepath, parameters, data)
+
+				data = GenerativeAD.Models.preprocess_images(data, parameters, denominator=32)
+				#(X_train,_), (X_val, y_val), (X_test, y_test) = data
+				training_info, results = fit(data, parameters)
+				# saving model separately
+				tagsave(joinpath(savepath, savename("model", parameters, "bson")), Dict("model"=>training_info.model), safe = true)
```
vitskvara: Check if model was trained at all, e.g. by `if training_info.model != nothing`

scripts/experiments_images/SkipGANomaly.jl
TheMrGhostman: done

scripts/experiments_images/SkipGANomaly.jl
TheMrGhostman: renamed

scripts/experiments_images/SkipGANomaly.jl
TheMrGhostman: only model saving or i should incoude the rest as well?

vitskvara:
Seems ok.

PR #25, opened 2020-09-15
Tabular datasets

vitskvara:


vitskvara:
This is ready, I am asking for a review

scripts/experiments_tabular/knn.jl
```diff
@@ -51,6 +51,7 @@ function fit(data, parameters)
 		global info, fit_t, _, _, _ = @timed fit!(model, data[1][1])
 	catch e
 		# return an empty array if fit fails so nothing is computed
+		rethrow(e)
```
janfrancu: I suppose this has been added to stop training of other seeds, if it happens to fail on one. Am I right?

src/datasets/tabular.jl
```diff
@@ -0,0 +1,383 @@
...
+	# I have opted for the original Loda datasets, use of multiclass problems in all vs one case
+	# does not necessarily represent a good anomaly detection scenario
+	data, _, _ = UCI.get_loda_data(dataset)
+	# return only easy and medium anomalies
+	UCI.normalize(data.normal, hcat(data.easy, data.medium)) # data (standardized)
+end
+
+other_datasets = ["annthyroid", "arrhythmia", "htru2", "kdd99", "kdd99_small", "spambase", "mammography", "har", "seismic"]
+function __init__()
+	register(
```
janfrancu, line 25: This is cool. :+1:

janfrancu (approved):
Good job. I will run my experiments too on the new data, once merged.

scripts/experiments_tabular/knn.jl
vitskvara: Actually this was part of debugging, it should not be there, because it will stop the training completely.

PR #28, opened 2020-09-16
Add RealNVP and MAF for tabular data.

### LINKED ISSUE #24: Implement general purpose RealNVP and MAF flows

janfrancu:
- [x] RealNVP

- [x] MAF

- [x] increase code reuse


janfrancu:
It should do the job. Ready for review.

scripts/experiments_tabular/maf.jl
```diff
@@ -0,0 +1,87 @@
...
+		global info, fit_t, _, _, _ = @timed fit!(model, data, parameters)
+ 	catch e
+		@info "Failed training due to \n$e"
+		return (fit_t = NaN,), []
+	end
+
+	training_info = (
+		fit_t = fit_t,
+		history = info.history,
+		model = nothing
```
vitskvara: Do you not save the model?

scripts/experiments_tabular/maf_run.sh
```diff
@@ -0,0 +1,17 @@
...
+#SBATCH --time=24:00:00
+#SBATCH --nodes=1 --ntasks-per-node=4 --cpus-per-task=1
+#SBATCH --mem=16G
+#SBATCH --qos==collaborator
+
+MAX_SEED=$1
+DATASET=$2
+
+module load Julia/1.5.1-linux-x86_64
+module load Python/3.8.2-GCCcore-9.3.0
```
vitskvara, line 11: Do you still need Python and py-env?

scripts/experiments_tabular/real_nvp.jl
```diff
@@ -0,0 +1,86 @@
...
+	model = GenerativeAD.Models.RealNVPFlow(
+				parameters.nflows,
+				D,
+				parameters.hsize,
+				parameters.nlayers)
+
+	try
+		global info, fit_t, _, _, _ = @timed fit!(model, data, parameters)
+ 	catch e
+		@info "Failed training due to \n$e"
```
vitskvara, line 42: nice

src/models/tabular_flows.jl
```diff
@@ -0,0 +1,115 @@
...
+	X = data[1][1]
+	X_val = data[2][1]
+
+	history = MVHistory()
+	patience = p.patience
+	reg = (p.wreg > 0) ? l2_reg : _ -> 0.0
+	
+	best_val_loss = loss(trn_model, X_val)
+	i = 1
+	for batch in RandomBatches(X, p.batchsize)
```
vitskvara, line 73: Does this mean that the loop runs indefinitely until convergence criterion is satisfied? What if it never is?

scripts/experiments_tabular/maf_run.sh
janfrancu: Unfortunatelly yes. I may be wrong, but once you import PyCall, it always needs some python to call home during startup. If we went the miniconda route, we would not have to deal with it.

scripts/experiments_tabular/maf.jl
janfrancu: Now that you mention it, I should.

scripts/experiments_tabular/maf.jl
```diff
@@ -0,0 +1,87 @@
...
+		)
+
+	training_info, [(x -> predict(info.model, x), parameters)]
```
vitskvara, line 55: Do you need to do `info.model`? Is not the original `model` variable changed during `fit!`? I think it is passed as reference.

scripts/experiments_tabular/maf.jl
vitskvara: Matěj now saves it in a separate file. Maybe we should consider it even for models where there is only one scoring function so it is consistent.

src/models/tabular_flows.jl
janfrancu: I though of giving it some upper limit like 5000, which is slightly above all the models that I have trained so far. Based on my observation 95% of models finish before that condition is met. These models like to overfit, so validation degrades reliably.

With all this said I will add some limit, in order to get the model before slurm kills after 24h, it would be wasteful.

scripts/experiments_tabular/maf.jl
janfrancu: That's the dirty checkpoint trick. I need to deepcopy the model, in order to have some backup. I don't know how I can assign the model back to it's original reference.

src/models/tabular_flows.jl
vitskvara: The time limit seems to make the most sense.

scripts/experiments_tabular/maf.jl
janfrancu: I will check it out.

scripts/experiments_tabular/maf.jl
vitskvara: You could copy the parameters from the last working version of train model to the original variable. But this works.

src/models/tabular_flows.jl
```diff
@@ -0,0 +1,115 @@
...
+	ps = Flux.params(trn_model);
+
+	X = data[1][1]
+	X_val = data[2][1]
+
+	history = MVHistory()
+	patience = p.patience
+	reg = (p.wreg > 0) ? l2_reg : _ -> 0.0
+	
+	best_val_loss = loss(trn_model, X_val)
```
vitskvara, line 70: Have you considered computing the loss over a whole validation dataset batch-wise? This might be very expensive for large datasets... or not? I usually do it on image data where only a batch of samples fits on the gpu.

vitskvara: What I mean is computing it batch-wise and then taking a mean.

vitskvara: I have tested it, and for elbo in a VAE model this starts to make sense from 5.000 samples (input dim = 4)

janfrancu: I will try to measure it with some fixed model architecture. I expect the time spent there is not insignificant in comparison to the whole training loop.

src/models/tabular_flows.jl
```diff
@@ -0,0 +1,123 @@
...
+				break
+			end
+		end
+		i += 1
+	end
+
+	# returning model in this way is not ideal
+	# it would have to modify the reference the
+	# underlying structure
+	(history=history, iterations=i, model=model)
```
vitskvara: Also return the number of trainable parameters (neurons).

src/models/utils/nn_builders.jl
```diff
@@ -0,0 +1,32 @@
...
+julia> build_mlp(4, 11, 1, 3, activation="relu", lastlayer="tanh")
+Chain(Dense(4, 11, relu), Dense(11, 11, relu), Dense(11, 1, tanh))
+```
+"""
+build_mlp(ks::Vector{Int}, fs::Vector) = Flux.Chain(map(i -> Dense(i[2],i[3],i[1]), zip(fs,ks[1:end-1],ks[2:end]))...)
+
+build_mlp(idim::Int, hdim::Int, odim::Int, nlayers::Int; kwargs...) =
+	build_mlp(vcat(idim, fill(hdim, nlayers-1)..., odim); kwargs...)
+
+function build_mlp(ks::Vector{Int}; activation::String = "relu", lastlayer::String = "")
```
vitskvara, line 25: Nice.

scripts/experiments_tabular/maf.jl
```diff
@@ -0,0 +1,92 @@
...
+				idim,
+				parameters.hdim,
+				parameters.nlayers,
+				parameters.ordering)
+
+	try
+		global info, fit_t, _, _, _ = @timed fit!(model, data, parameters)
+ 	catch e
+		@info "Failed training due to \n$e"
+		return (fit_t = NaN,), []
```
vitskvara: Consider setting the remaining fields as well, e.g. `model = nothing`, see my comment on model saving.

scripts/experiments_tabular/maf.jl
```diff
@@ -0,0 +1,92 @@
...
+		edited_parameters = GenerativeAD.edit_params(data, parameters)
+
+		if GenerativeAD.check_params(savepath, data, edited_parameters)
+			@info "Started training $(modelname)$(edited_parameters) on $(dataset):$(seed)"
+			@info "Train/valdiation/test splits: $(size(data[1][1], 2)) | $(size(data[2][1], 2)) | $(size(data[2][1], 2))"
+			@info "Number of features: $(size(data[1][1], 1))"
+			
+			training_info, results = fit(data, edited_parameters)
+
+			tagsave(joinpath(savepath, savename("model", parameters, "bson")), Dict("model"=>training_info.model), safe = true)
```
vitskvara: This is gonna fail in case of error in fit, since there is no field model. Consider checking `if training_info.model != nothing`

scripts/experiments_tabular/maf.jl
```diff
@@ -0,0 +1,92 @@
...
+			training_info = [p for p in pairs(training_info) if p[1] != :model]
```
vitskvara: `merge(training_info, (model = nothing,))` is cleaner

scripts/experiments_tabular/maf.jl
janfrancu: Good catch.

PR #31, opened 2020-09-23
VAE

vitskvara:


scripts/evaluate_performance.jl
```diff
@@ -42,7 +43,7 @@ function query_stats(target::String)
 			@info "$target not compatible"
 		end
 	else
-		query_stats.(joinpath.(target, readdir(target)))
+		query_stats.(joinpath.(target, filter(x->!(occursin("model_", string(x))), readdir(target))))
```
janfrancu, line 46: I have noticed this earlier, thanks for including it in the PR.

scripts/experiments_tabular/vae.jl
```diff
@@ -0,0 +1,142 @@
...
+	sample_params()
+
+Should return a named tuple that contains a sample of model parameters.
+"""
+function sample_params()
+	par_vec = (2 .^(3:8), 2 .^(4:9), 10f0 .^(-4:-3), 2 .^ (5:7), ["relu", "swish", "tanh"], 3:4, 1:Int(1e8))
+	argnames = (:zdim, :hdim, :lr, :batchsize, :activation, :nlayers, :init_seed)
+	parameters = (;zip(argnames, map(x->sample(x, 1)[1], par_vec))...)
+	# ensure that zdim < hdim
+	while parameters.zdim >= parameters.hdim
```
janfrancu, line 43: What about putting this into `edit_params`, where you can divide by 2 until the condition is satisfied?

scripts/experiments_tabular/vae.jl
```diff
@@ -0,0 +1,142 @@
...
+Each element of the return vector contains a specific anomaly score function - there can be multiple for each trained model.
+Final parameters is a named tuple of names and parameter values that are used for creation of the savefile name.
+"""
+function fit(data, parameters)
+	# construct model - constructor should only accept kwargs
+	model = GenerativeAD.Models.vae_constructor(;idim=size(data[1][1],1), parameters...)
+
+	# fit train data
+	try
+		global info, fit_t, _, _, _ = @timed fit!(model, data; max_train_time=82800,
```
janfrancu: I realized yesterday, that the maximum training time of 23 hours does not take into account, that this script runs 5 trainings. I don't know if we should limit the training to 23/5 hours or just leave it be.

scripts/experiments_tabular/vae.jl
```diff
@@ -0,0 +1,142 @@
...
+	# now return the different scoring functions
+	training_info, [
+		(x -> GenerativeAD.Models.reconstruction_score(info.model, x), merge(parameters, (score = "reconstruction",))),
+		(x -> GenerativeAD.Models.latent_score(info.model, x), merge(parameters, (score = "latent",))),
+		]
+end
+function GenerativeAD.edit_params(data, parameters)
+	idim = size(data[1][1],1)
+	# put the largest possible zdim where zdim < idim, the model tends to converge poorly if the latent dim is larger than idim
+	if parameters.zdim >= idim
```
janfrancu, line 87: Is this duplicate of the code in `sample_params`?

src/models/vae.jl
```diff
@@ -0,0 +1,150 @@
...
+	(nlayers < 2) ? error("Less than 3 layers are not supported") : nothing
+	# function from string
+	act = eval(Meta.parse(activation))
+
+	# if seed is given, set it
+	(init_seed != nothing) ? Random.seed!(init_seed) : nothing
+	
+	# construct the model
+	# encoder - diagonal covariance
+	encoder_map = Chain(
```
janfrancu, line 29: I see that you did not use the `build_mlp` methods. Is there something missing in their implementation?

janfrancu:
```julia
decoder_map = Chain(
		build_mlp(idim, hdim, hdim, nlayers-1, activation=activation, lastlayer="linear")...,
		ConditionalDists.SplitLayer(hdim, [zdim, zdim], [identity, safe_softplus])
		)
```
Should be equivalent, but I have not checked it.

src/models/vae.jl
```diff
@@ -0,0 +1,150 @@
...
+
+"""
+	loss(model::GenerativeModels.VAE, x[, batchsize])
+
+Negative ELBO for training of a VAE model.
+"""
+loss(model::GenerativeModels.VAE, x) = -elbo(model, x)
+# version of loss for large datasets where
+loss(model::GenerativeModels.VAE, x, batchsize::Int) = 
+	mean(map(y->loss(model,y), Flux.Data.DataLoader(x, batchsize=batchsize)))
```
janfrancu, line 57: Nice use of dispatch.

src/models/vae.jl
```diff
@@ -0,0 +1,150 @@
...
+		else # else stop if the model has not improved for `patience` iterations
+			_patience -= 1
+			if _patience == 0
+				@info "Stopped training after $(i) iterations."
+				break
+			end
+		end
+		i += 1
+	end
+	# again, this is not optimal, the model should be passed by reference and only teh reference should be edited
```
janfrancu: spelling *the

src/models/vae.jl
```diff
@@ -0,0 +1,150 @@
...
+		batch_loss = 0f0
+		gs = gradient(() -> begin 
+			batch_loss = loss(tr_model,batch)
+		end, ps)
+	 	Flux.update!(opt, ps, gs)
+
+		# validation/early stopping
+		val_loss = (val_N > 5000) ? loss(tr_model, val_x, 256) : loss(tr_model, val_x)
+		
+		(i%check_interval == 0) ? (@info "$i - loss: $(batch_loss) (batch) | $(val_loss) (validation)") : nothing
```
janfrancu, line 90: Is `check_interval` is used only for printing or did you intend to check also the validation loss also less often? I'm asking just if it makes sense to put it as a parameter, because currently it's just a debugging technicality.

src/models/vae.jl
```diff
@@ -0,0 +1,150 @@
...
+	reconstruct(model::GenerativeModels.VAE, x)
+
+Data reconstruction.
+"""
+reconstruct(model::GenerativeModels.VAE, x) = mean(model.decoder, rand(model.encoder, x))
+
+"""
+	reconstruction_score(model::GenerativeModels.VAE, x)
+
+Anomaly score based on
```
janfrancu: Finish the description.

src/models/vae.jl
```diff
@@ -0,0 +1,150 @@
...
+	p = condition(model.decoder, rand(model.encoder, x))
+	-logpdf(p, x)
+end
+"""
+	latent_score(model::GenerativeModels.VAE, x) 
+"""
+function latent_score(model::GenerativeModels.VAE, x) 
+	z = rand(model.encoder, x)
+	-logpdf(model.prior, z)
+end
```
janfrancu: What about the `predict` function abstraction from StatsBase. Is it possible to use it here?

scripts/experiments_tabular/vae.jl
vitskvara: I don't like dividing it by two since then it is truly random anymore - higher values closer to hdim will tend to get selected more often. Also, I would have to redefine the sampling range for the parameter again in `edit_params` which I want to avoid since it would be easy to edit it in once place and not in the other.

scripts/experiments_tabular/vae.jl
vitskvara: That's a good question. I would do it for 23/5 hours, since then some models will never finish. I will change the input value when the `fit!` function is called in `fit`.

scripts/experiments_tabular/vae.jl
vitskvara: Nope. In `sample_params`, I want to ensure that zdim < hdim, independently on data. Here, I want to ensure that zdim < idim, so that I cant learn identity.

src/models/vae.jl
vitskvara: I wanted to use it but did not get to it - thanks!

src/models/vae.jl
vitskvara: Thanks.

src/models/vae.jl
vitskvara: Good catch.

src/models/vae.jl
vitskvara: Yeah it's only used for printing so that there is less output. I can see it being an argument of the function for debugging of model, not really a parameter to be sampled from.

src/models/vae.jl
vitskvara: Thanks

src/models/vae.jl
vitskvara: Yeah, since there are two different scores, I don't see the necessity for overloading of `predict`.

scripts/experiments_tabular/vae.jl
janfrancu: I see. From afar it looked really similar.

scripts/experiments_tabular/vae.jl
janfrancu: Makes sense.

src/models/vae.jl
janfrancu: Ok.

janfrancu (approved):

PR #34, opened 2020-09-25
WAE

vitskvara:
Wasserstein autoencoder etc.


scripts/experiments_tabular/wae.jl
```diff
@@ -0,0 +1,164 @@
...
+    "dataset"
+        required = true
+        arg_type = String
+        help = "dataset"
+end
+# for testing purposes
+if length(ARGS) == 0
+	push!(ARGS, "1")
+	push!(ARGS, "iris")
+end
```
janfrancu, line 28: I would leave this out, even though it does nothing when run in batch with some arguments.

scripts/experiments_tabular/wae.jl
```diff
@@ -0,0 +1,164 @@
...
+	end
+	loss(m,x) = parameters.lambda*mmd_mean(m, x, k) .- mean(logpdf(m.decoder, x, rand(m.encoder, x)))
+
+	# fit train data
+	try
+		global info, fit_t, _, _, _ = @timed fit!(model, data, loss; max_train_time=82800/max_seed, 
+			patience=200, check_interval=10, parameters...)
+	catch e
+		# return an empty array if fit fails so nothing is computed
+		rethrow(e)
```
janfrancu: You have probably forgotten to remove this one.

src/models/vae.jl
```diff
@@ -16,9 +16,13 @@ safe_softplus(x::T) where T  = softplus(x) + T(0.000001)
 
 
 """
-	vae_constructor(;idim::Int=1, zdim::Int=1, activation = "relu", hdim=128, nlayers=3, kwargs...)
+	vae_constructor(;;idim::Int=1, zdim::Int=1, activation = "relu", hdim=128, nlayers::Int=3, 
+		init_seed=nothing, prior="normal", pseudoinput_mean=nothing, k=1, kwargs...)
+
+Constructs a classical variational autoencoder.
```
janfrancu, line 22: Add comment about the additional parameters as it is no longer supporting only classical VAE.

janfrancu (changes requested):

scripts/experiments_tabular/wae.jl
vitskvara: Yes, thanks for the catch

src/models/vae.jl
vitskvara: Good idea.

scripts/experiments_tabular/wae.jl
vitskvara: I'd like to keep it for cases when I test this in REPL.

vitskvara: Also, it is needed in order to avoid errors when including the script elsewhere.

janfrancu: It can be achieved much cleanly using default argument values. 
```
@add_arg_table! s begin
   "max_seed"
        default = 1
        arg_type = Int
        help = "seed"
    "dataset"
        default = "iris"
        arg_type = String
        help = "dataset"
end
```
Have not tested, but I remember using it this way.

janfrancu (approved):

PR #35, opened 2020-09-25
Check params fix

vitskvara:
Now it should work even if the parameter names have changed. Also, removed unnecessary argument `data` of the function.

### DIFF
--- a/src/experiments.jl
+++ b/src/experiments.jl
@@ -44,11 +44,11 @@ function edit_params(data, parameters)
 end
 
 """
-	check_params(edit_params_f, savepath, data, parameters)
+	check_params(savepath, parameters)
 
-This checks if the model with given parameters wasn't already trained and saved. 
+Returns `true` if the model with given parameters wasn't already trained and saved. 
 """
-function check_params(savepath, data, parameters)
+function check_params(savepath, parameters)
 	if ~isdir(savepath)
 		return true
 	end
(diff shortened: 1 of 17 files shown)


test/experiments.jl
```diff
@@ -0,0 +1,15 @@
+@testset "Experiments" begin
+	using DrWatson
+	@quickactivate
+	using Test
+	using GenerativeAD
+
+	tstpath = joinpath(dirname(pathof(GenerativeAD)), "../test/misc")
```
janfrancu: Creating the files on demand during test would make this more transparent.

src/experiments.jl
```diff
@@ -58,8 +58,11 @@ function check_params(savepath, data, parameters)
 	fs = filter(x->!(startswith(x, "model")), fs)
 	# if the first argument name contains a "_", than the savename is parsed wrongly
 	saved_params = map(x -> DrWatson.parse_savename("_"*x)[2], fs)
+	# now filter out saved models where parameter names are different or missing
+	pkeys = collect(keys(parameters))
+	filter!(ps->intersect(pkeys, Symbol.(collect(keys(ps))))==pkeys, saved_params)
 	for params in saved_params
-		all(map(k->params[String(k)] == parameters[k], collect(keys(parameters)))) ? (return false) : nothing
```
janfrancu, line 62: I've had in mind something, that will ignore skip those files that do not match the same key names. This is however trying to match the values on the intersection of names, which I don't really consider as match as they may come from some weird version of the model.

Oh I see now, that what if the intersection is not equal to pkeys, it will skip the for cycle and go straight to the end. Leaving the above just as a note, that the code is not as clear even for someone who wrote the issue.

test/experiments.jl
vitskvara: I thought about that but was worried about file permissions... thinking again they should probably not be a problem. Will redo this.

src/experiments.jl
vitskvara: Yes, the added line is supposed to filter out models where some of the new parameter names are missing.

PR #37, opened 2020-09-29
Better evaluation

### LINKED ISSUE #10: Evaluation
Write a proper evaluation framework. A set of models should result in a DataFrame that contains the model name, dataset, seed, hyperparameters and a few columns of evaluation metrics on validation and test data. This is then saved as a csv or 2 csvs (for validation and test). Finally, we should have a function that collects the results.

janfrancu:
- [x] initial prototype from existing script

- [x] move common code to src

- [x] test properly in order not to overwrite existing experimental results

- [x] create simple scripts to regenerate results and print basic tables


scripts/evaluate_performance.jl
```diff
@@ -1,24 +1,52 @@
...
+		row = merge(row, (anomaly_class = r[:anomaly_class]))
+	end
 
-	results = []
-	for (scores, labels) in scores_labels
+	for splt in ["val", "tst"]
+		scores = r[_prefix_symbol(splt, :scores)]
+		labels = r[_prefix_symbol(splt, :labels)]
+
+		any(isnan.(scores)) && error("$(splt)_scores contain NaN")
```
vitskvara: Can we filter out the scores and labels that contain NaN?

janfrancu: Yes we can, however I would not go as far as filtering out all NaNs as oftentimes you are left with just an empty array. I am not sure if we should tolerate some NaNs given some threshold. Are 4 NaNs a sign of corrupted model/results or is it 5? Who is here to say, that the methods just does not purposefully fail on so many occasions such that the performance seems better.
This is maybe too harsh of a limit as some of the methods such as `ABOD` on `kdd, mammography, miniboone` produce at least on NaN for each hyperparameter, yielding no evaluation files for this dataset/model combination.

vitskvara: I would put the limit probably at one half.

scripts/evaluate_performance.jl
```diff
@@ -29,22 +57,163 @@ function compute_stats(f::String)
...
+Chooses agregates eval metrics by seed over a given hyperparameter and then chooses best
+model based on `criterion_col`. output_cols 
+"""
+function aggregate_stats(df::DataFrame, criterion_col=:val_auc, output_cols=[:tst_auc])
+	metrics = ["auc", "auprc", "tpr@5", "f1@5"]
+	agg_cols = vcat(_prefix_symbol.("val", metrics), _prefix_symbol.("tst", metrics))
+
+	# agregate by seed over given hyperparameter and then choose best
+	results = []
+	for (dkey, dg) in pairs(groupby(df, :dataset))
```
vitskvara: For image data, there has to be another aggregation over classes.

janfrancu:
This is as far as I will push it in this pull request. It's getting quite bloated and disconnected.

scripts/evaluation/README.md
```diff
@@ -0,0 +1,45 @@
+# Evaluation scripts
+Due to the large number of files that need to be processed the main evaluation is split into three steps
+- generation of files with statistics (parallel) - `generate_stats.jl`
+- collection of previously generated files into one cached dataframe (parallel) - `collect_stats.jl`
+- printing of summary tables - `evaluate_performance.jl`
+
+The first two steps can be run as a batch job by running `sbatch run_eval.sh` script. By default this won't rewrite any precomputed files with the exception of the cached DataFrame in the second step.
```
vitskvara: does this mean that the first script saves copies of the files with statistics or it skips them if they exist?

scripts/evaluation/README.md
```diff
@@ -0,0 +1,45 @@
...
+- loading `evaluation/images_eval.bson` cache, using `val_pat_x` with increasing `x` for sorting models and `tst_pat_10` for final ranking, printing the ranking given percentage of labeled samples in the validation to txt file in data prefixed folder given by `--output-prefix`
+```
+    julia --project ./evaluate_performance.jl \
+                        evaluation/images_eval.bson \
+                        --output-prefix evaluation/images_eval \
+                        --rank-metric tst_pat_10 \
+                        --proportional
+```
+
+Some combination of parameters don't make sense, such as running with `--best-params` while also using `--proportional`. Furthermore the tex(latex) output does not escape underscores and therefore cannot be parsed sometimes. The code scaling to multiple threads does not work optimally, but it is encouraged to use 32 threads when processing more than 50k files.
```
vitskvara: Nice, I like all the possibilities. Underscore escaping should be fixable, right?

scripts/evaluation/collect_stats.jl
```diff
@@ -0,0 +1,56 @@
...
+using FileIO
+using DataFrames
+using Base.Threads: @threads
+using GenerativeAD
+
+s = ArgParseSettings()
+@add_arg_table! s begin
+    "source_prefix"
+		arg_type = String
+		default = "evaluation/images"
```
vitskvara, line 15: Is it wise to use default values here? Can someone unknowingly overwrite something by running this accidentally?

scripts/evaluate_performance_single.jl
```diff
@@ -1,4 +1,4 @@
-# run this script as `julia evaluate_performance.jl res.bson` or `julia evaluate_performance.jl dir`
+# run this script as `julia evaluate_performance_single.jl res.bson` or `julia evaluate_performance_single.jl dir`
```
vitskvara, line 1: There is no code reuse? Everything is as it was?

src/evaluation/utils.jl
```diff
@@ -0,0 +1,18 @@
+function collect_files!(target::String, files)
+	if isfile(target)
+		push!(files, target)
+	else
+		for file in readdir(target, join=true)
```
vitskvara, line 5: did not know you can use the join kwarg to avoid joinpath, nice

scripts/evaluation/run_eval.sh
```diff
@@ -0,0 +1,26 @@
...
+#SBATCH --qos==collaborator
+
+module load Julia/1.5.1-linux-x86_64
+module load Python/3.8.2-GCCcore-9.3.0
+
+# load virtualenv
+source ${HOME}/julia-env/bin/activate
+export PYTHON="${HOME}/julia-env/bin/python"
+
+julia --threads 32 --project ./generate_stats.jl experiments/images evaluation/images
```
vitskvara, line 19: Again, does this skip models for which the evaluation already exists?

scripts/evaluation/generate_stats.jl
```diff
@@ -0,0 +1,75 @@
...
+	
+	source = datadir(source_prefix)
+	@info "Collecting files from $source folder."
+	files = GenerativeAD.Evaluation.collect_files(source)
+	# filter out model files
+	filter!(x -> !startswith(basename(x), "model"), files)
+	@info "Collected $(length(files)) files from $source folder."
+	# it might happen that when appending results some of the cores
+	# just go over already computed files
+	files = files[randperm(length(files))]
```
vitskvara, line 50: I dont understand this.

scripts/evaluation/generate_stats.jl
```diff
@@ -0,0 +1,75 @@
...
+
+	@threads for f in files
+		try
+			target_dir = dirname(replace(f, source_prefix => target_prefix))
+			target = joinpath(target_dir, "eval_$(basename(f))")
+			if (isfile(target) && force) || ~isfile(target)
```
vitskvara, line 56: Ok I see now how it works with overwriting files.

src/evaluation/stats.jl
```diff
@@ -0,0 +1,142 @@
...
+_prefix_symbol(prefix, s) = Symbol("$(prefix)_$(s)")
+
+
+"""
+	_precision_at(p, scores, labels)
+
+Computes precision on portion `p` samples with highest score.
+"""
+function _precision_at(p, labels, scores)
+	pN = Int(round(floor(p*length(labels)))) # round is more forgiving
```
vitskvara: Here you can do just `pN = floor(Int, p*length(labels))`

src/evaluation/stats.jl
```diff
@@ -0,0 +1,142 @@
...
+Returns a DataFrame row with metrics and additional metadata for groupby's.
+Hash of the model parameters is precomputed in order to make the groupby easier.
+"""
+function compute_stats(f::String)
+	r = load(f)
+	row = (
+		modelname = r[:modelname],
+		dataset = r[:dataset],
+		phash = hash(r[:parameters]),
+		parameters = savename(r[:parameters], digits=6),
```
vitskvara, line 51: It's a good idea to compare models with the hash, since savename ignores array arguments.

src/evaluation/stats.jl
```diff
@@ -0,0 +1,142 @@
...
+		labels = r[_prefix_symbol(splt, :labels)]
+
+		if length(scores) > 1
+			invalid = isnan.(scores)
+			ninvalid = sum(invalid)
+
+			if ninvalid > 0
+				invrat = ninvalid/length(scores)
+				invlab = labels[invalid]
+				cml = countmap(invlab)
```
vitskvara, line 69: did not know `countmap`, thats useful

vitskvara (approved):

scripts/evaluation/README.md
janfrancu: Definitely on my todo list, but there are more important issues now.

scripts/evaluate_performance_single.jl
janfrancu: I though of bringing the common code from GenerativeAD, but then the script would not be as responsive as it is now.

scripts/evaluation/run_eval.sh
janfrancu: Yes, by default it skips them. If we were to add some new metric or change the generation in some significant way, I would however still prefer to delete all previous files and regenerate them without the force flag, as one cannot be sure that after the script there will be only one version of the files.

scripts/evaluation/collect_stats.jl
janfrancu: Running the collection accidentally should just update the default dataframe, which serves as a cache for the current state of computed experiments. If you intend to change those dataframes, for example group some models together, etc. I would advise you to save it in a different one, so that the default name corresponds to this raw cache.

scripts/evaluation/generate_stats.jl
janfrancu: My suspicion is that when assigning the work to different threads, the scheduler splits the iterator into chunks and does not schedule new work once a thread completes. In case of adding files for new experiments this mean that only some of the threads were seem to be running as the rest just went through their share of files and confirmed that they exist and there is no need to compute anything. This is probably not the case, as it seems rather silly way of multi-threading, however from the CPU utilization graph it seemed to behave like that.

src/evaluation/stats.jl
janfrancu: Thanks for spotting this. It is an artifact from when I was playing with both floor and round options.

src/evaluation/stats.jl
janfrancu: Right!? In python I so got to used to pandas's value counts (which worked only on their series) and was really glad that they added something like that to Julia.

scripts/evaluate_performance_single.jl
vitskvara: That seems sensible

scripts/evaluation/generate_stats.jl
vitskvara: Ah, now I understand. Thanks for the explanation.

PR #38, opened 2020-10-01
AAE

vitskvara:
Adversarial autoencoders.


src/models/aae.jl
```diff
@@ -0,0 +1,282 @@
...
+end
+
+"""
+	AAE(prior, encoder, decoder, discriminator)
+	AAE(zdim::Int, encoder, decoder, discriminator)
+
+Constructs an adversarial autoencoder.
+"""
+function AAE(zdim::Int, encoder, decoder, discriminator)
+	W = first(Flux.params(encoder))
```
janfrancu, line 15: Align.

src/models/vae.jl
```diff
@@ -14,6 +16,16 @@ Safe version of softplus.
...
 
+"""
+	init_vamp(pseudoinput_mean,k::Int)
+
+Initializes the VAMP prior from a mean vector and number of components.
+"""
+function init_vamp(pseudoinput_mean, k::Int)
+	T = eltype(pseudoinput_mean)
+	pseudoinputs = T(1) .* randn(T, size(pseudoinput_mean)[1:end-1]..., k) .+ pseudoinput_mean
+	VAMP(pseudoinputs)
```
janfrancu, line 27: Out of curiosity. Where is VAMP imported from?

src/models/aae.jl
```diff
@@ -0,0 +1,282 @@
...
+#loss(m::AAE,x,batchsize::Int) = aeloss(m,x,batchsize) + dloss(m,x,batchsize) + gloss(m,x,batchsize)
+
+"""
+	StatsBase.fit!(model::AAE, data::Tuple; max_train_time=82800, 
+	lr=0.001, batchsize=64, patience=30, check_interval::Int=10, kwargs...)
+"""
+function StatsBase.fit!(model::AAE, data::Tuple; max_train_time=82800, lr=0.001, 
+	batchsize=64, patience=30, check_interval::Int=10, kwargs...)
+	history = MVHistory()
+	aeopt = ADAM(lr)
```
janfrancu, line 148: Why do you have more than one optimizer, if they just share the same learning rate?

src/models/aae.jl
```diff
@@ -0,0 +1,282 @@
...
+			_patience = patience
+
+			# this should save the model at least once
+			# when the validation loss is decreasing 
+			if mod(i, 10) == 0
+				model = deepcopy(tr_model)
+			end
+		elseif time() - start_time > max_train_time # stop early if time is running out
+			model = deepcopy(tr_model)
+			@info "Stopped training after $(i) iterations, $((time() - start_time)/3600) hours."
```
janfrancu, line 217: Good idea to print the time it took.

src/models/aae.jl
```diff
@@ -0,0 +1,282 @@
...
+	# again, this is not optimal, the model should be passed by reference and only the reference should be edited
+	(history=history, iterations=i, model=model, npars=sum(map(p->length(p), Flux.params(model))))
+end
+
+"""
+	AEModel
+
+A Union of VAE and AAE types.
+"""
+AEModel = Union{VAE, AAE}
```
janfrancu, line 237: This may be not the best place for the shared code, but let's leave it here.

janfrancu (changes requested):

src/models/aae.jl
vitskvara: Alright

src/models/vae.jl
vitskvara: It's from GenerativeModels.jl

src/models/aae.jl
vitskvara: Because each optimizer optimizes different parts of the model and different set of parameters, therefore the history (moments etc.) is different. Or can this be achieved by using a single optimizer for all parameters?

src/models/aae.jl
vitskvara: Yeah it's not really a lot of code so I thought it does not really need a separate file.

src/models/aae.jl
janfrancu: Don't know. 
If it works, don't change it. :smile:

janfrancu (approved):

src/models/aae.jl
vitskvara: originally, I only had one optimizer and was surprised that it worked at all - maybe it is restarted everytime it encounters new parameters? But the convergence is much better with separate optimizers.

PR #39, opened 2020-10-05
GANomaly on tabular data

TheMrGhostman:
Ready to be cheched


scripts/experiments_images/GANomaly.jl
```diff
@@ -77,11 +78,11 @@ function fit(data, parameters)
...
 		model = (info[2]|>cpu, info[3]|>cpu),
 		history = info[1], # losses through time
-        npars = info[4], # number of parameters
-        iters = info[5] # optim iterations of model
+		npars = info[4], # number of parameters
+		iters = info[5] # optim iterations of model
 		)
 
-	return training_info, [(x -> GenerativeAD.Models.anomaly_score_gpu(generator|>cpu, x; dims=3), parameters)]
+	return training_info, [(x -> GenerativeAD.Models.anomaly_score_gpu(generator|>cpu, x; dims=3)[:], parameters)]
```
vitskvara, line 85: Is this a cpu or gpu array?

scripts/experiments_tabular/GANomaly.jl
```diff
@@ -0,0 +1,115 @@
...
+    par_vec = (
+        2 .^(4:9),
+        2 .^(3:8),
+        3:4,
+        ["relu", "swish", "tanh"],
+        10f0 .^ (-4:-3),
+        2 .^ (5:7),
+        [10000],
+        [30],
+        [10],
```
vitskvara, line 50: Is this enough? I have set patience to 200 in order to get out of certain local extremas.

scripts/experiments_images/GANomaly.jl
TheMrGhostman: `|> cpu` is probably useless here. The function is written to work on GPU (with limited memory -> batches), but when computer does not have GPU it runs on cpu instead. 

but it always returns cpu array

vitskvara (commented):
Make sure to use the proper toml files.

scripts/experiments_tabular/GANomaly.jl
TheMrGhostman: i am checking every 30 iterations with patience "10" => 30*10=300 iterations patience

TheMrGhostman: but i can change it to check_every = 1 and patience = 200 than it will be the same as yours

TheMrGhostman:
> Make sure to use the proper toml files.

should be ok now

vitskvara:
> > Make sure to use the proper toml files.
> 
> should be ok now

You still seem to be using the old version of Manifest.toml with lot of old versions of dependencies.

TheMrGhostman:
> > > Make sure to use the proper toml files.
> > 
> > 
> > should be ok now
> 
> You still seem to be using the old version of Manifest.toml with lot of old versions of dependencies.

it is manifest.toml downloaded from current master

scripts/experiments_tabular/GANomaly.jl
vitskvara: ok. nevermind then

vitskvara:
I don't see any changesto Manifest in the latest commit.

TheMrGhostman:
that is weird. I copied it again from master but it seems it is the same file. Git see no changes

PR #42, opened 2020-10-07
MO_GAAL

TheMrGhostman:
- adding MO_GAAL
- adding weight decay


vitskvara (approved):

vitskvara:
Looks ok, try to test the run in a local version of the repo before merging and pulling.

TheMrGhostman:
ok

vitskvara:
One last thing - can you somehow obtain and store the number of tunable parameters/weights of the model?

TheMrGhostman:
> One last thing - can you somehow obtain and store the number of tunable parameters/weights of the model?

i think it is there ... 	https://github.com/aicenter/GenerativeAD.jl/blob/50571c1e1da599ad97d5b309289c563bdc418b68/src/models/GANomaly.jl#L365
i store it in save entries https://github.com/aicenter/GenerativeAD.jl/blob/50571c1e1da599ad97d5b309289c563bdc418b68/scripts/experiments_images/GANomaly.jl#L81

vitskvara:
Yes, that works for GANomaly, but I mean parameters of MOGAAL.

TheMrGhostman:
> Yes, that works for GANomaly, but I mean parameters of MOGAAL.

oh, mogaal. sure i can add python function there.

TheMrGhostman:
when i think about it now. For mogaal there is fixed number of weights and that number only depends on dimensions of data.

vitskvara:
Also on the number of discriminators.

vitskvara:
I think it should be possible to compute it from the input parameters externally, without the need to add any python code.

TheMrGhostman:
that is what i meant in second reply. I am already working on it

scripts/experiments_tabular/mogaal_run.sh
```diff
@@ -6,12 +6,12 @@
...
 DATASET=$2
 
-module load Julia/1.4.1-linux-x86_64
+module load Julia/1.5.1-linux-x86_64
 module load Python/3.8.2-GCCcore-9.3.0
 
 # load virtualenv
-source ${HOME}/sklearn-env/bin/activate
-export PYTHON="${HOME}/sklearn-env/bin/python"
+source ${HOME}/ad/bin/activate
```
vitskvara, line 13: This is potentially dangerous if you run two scripts that are using different python environments - PyCall is gonna be compiled over and over again and things might crash.

PR #43, opened 2020-10-08
GAN

vitskvara:


vitskvara:
I implemented the WGAN model as well, but without run scripts, as it does not seem to be applicable to AD.

src/models/gan.jl
```diff
@@ -0,0 +1,191 @@
...
+using Distributions
+using DistributionsAD
+using StatsBase
+using Random
+
+"""
+	gan_constructor(;idim::Int=1, zdim::Int=1, activation = "relu", hdim=128, nlayers::Int=3, 
+		init_seed=nothing, prior="normal", pseudoinput_mean=nothing, k=1, kwargs...)
+
+Constructs a classical variational autoencoder.
```
janfrancu: Incorrect docs.

scripts/count_runs.jl
```diff
@@ -0,0 +1,58 @@
...
+
+function list_files(datapath)
+	fs = readdir(datapath)
+	fs = filter(f->!(occursin("model_", f)), fs)
+	fs = filter(f->!(occursin("#", f)), fs)
+	return unique(fs)
+end
+function _check_validity(file)
+	data = load(file)
+	return (length(data[:val_labels]) == length(data[:val_scores])) && length(data[:tst_labels]) == length(data[:tst_scores])
```
janfrancu: Checking `any(isnan.(___))` may be also useful.

vitskvara: True, will add that.

src/models/gan.jl
vitskvara: Thanks.

PR #45, opened 2020-10-12
[WIP] conv nets

vitskvara:


vitskvara:
There is a working VAE, I'd like to start running it asap, so I'd like to merge it and then continue adding more models.

src/models/utils/vae_utils.jl
```diff
@@ -0,0 +1,102 @@
...
+"""
+encode_mean(model::AEModel, x) = mean(model.encoder, x)
+"""
+	encode_mean_gpu(model::AEModel, x[, batchsize])
+
+Produce data encodings. Works only on 4D tensors.
+"""
+encode_mean_gpu(model, x::AbstractArray{T,4}) where T = encode_mean(model, gpu(Array(x)))
+function encode_mean_gpu(model, x::AbstractArray{T,4}, batchsize::Int) where T
+	# this has to be done in a loop since doing cat(map()...) fails if there are too many arguments
```
janfrancu, line 55: Prealocating is always better when manipulating large data, even though it may not be the nicest looking solution.
There are some optimized implementations of `reduce(cat, iter)`, `reduce(hcat, )`, that allocate once, however one cannot be always certain that they will be called.

src/models/vae.jl
```diff
@@ -81,6 +81,68 @@ function vae_constructor(;idim::Int=1, zdim::Int=1, activation = "relu", hdim=12
...
+- `channels=(1,1)`: number of channels.
+- `scalings=(1,1)`: scalings in subsewuent layers.
+- `init_seed=nothing`: seed to initialize weights.
+- `prior="normal"`: one of ["normal", "vamp"].
+- `pseudoinput_mean=nothing`: mean of data used to initialize the VAMP prior.
+- `k::Int=1`: number of VAMP components.
+- `batchnorm=false`: use batchnorm (discouraged). 
+"""
+function conv_vae_constructor(;idim=(2,2,1), zdim::Int=1, activation="relu", hdim=1024, 
+	kernelsizes=(1,1), channels=(1,1), scalings=(1,1),
```
janfrancu, line 106: Indent please, so it's clear that it comes from the function header.

janfrancu (approved):

src/models/vae.jl
vitskvara: Done

PR #46, opened 2020-10-15
adVAE

TheMrGhostman:


scripts/experiments_images/adVAE.jl
```diff
@@ -0,0 +1,134 @@
...
+        :extra_layers,
+        :gamma,
+        :lambda,
+        :mx,
+        :mz,
+        :lr, 
+        :decay,
+        :batch_size, 
+        :iters, 
+        :check_every,
```
vitskvara: I would not put theses parameters here as they never change and make the savename too long. You can just put them in the constructor or fit function directly.

scripts/experiments_images/adVAE.jl
```diff
@@ -0,0 +1,134 @@
...
+
+function fit(data, parameters)
+	# define models (Generator, Discriminator)
+	advae = adVAE(;parameters...)
+
+	# define optimiser
+	try
+		global info, fit_t, _, _, _ = @timed fit!(advae |> gpu, data, parameters)
+	catch e
+		println("Error caught.")
```
vitskvara: Maybe print the error message?

scripts/experiments_images/adVAE.jl
```diff
@@ -0,0 +1,134 @@
...
+        [30],
+        1:Int(1e8),
+    )
+
+    return NamedTuple{argnames}(map(x->sample(x,1)[1], par_vec))
+end
+
+function fit(data, parameters)
+	# define models (Generator, Discriminator)
+	advae = adVAE(;parameters...)
```
vitskvara: conv_advae?

scripts/experiments_images/adVAE.jl
```diff
@@ -0,0 +1,134 @@
...
+		required = true
+		arg_type = String
+		help = "dataset"
+end
+parsed_args = parse_args(ARGS, s)
+@unpack dataset, max_seed = parsed_args
+
+modelname ="adVAE"
+
+function sample_params()
```
vitskvara, line 32: are these parameters for conv advae?

src/models/GANomaly.jl
```diff
@@ -250,6 +250,8 @@ end
 computes generator and discriminator loss without additional pass through model
 """
 function validation_loss(g::Generator, d::Discriminator, real_input; weights=[1,50,1])
```
vitskvara: I dont like the fact that switching to and from testmode is done here, it is very untransparent. Also, if the generator/discriminator is in testmode in the first place, after calling this function, it is not and the user can't possibly know it.

vitskvara: I would prefer the switching to be done outside of the function.

src/models/GANomaly.jl
```diff
@@ -271,18 +274,22 @@ end
 computes unscaled anomaly score A(x) = || E1(x) - E2(D(E1(x))) ||_1
 """
 function anomaly_score(generator::Generator, real_input; dims=3)
```
vitskvara: The same as above, dont do testmode switching inside the function.

src/models/SkipGANomaly.jl
```diff
@@ -185,6 +185,7 @@ end
 computes generator and discriminator loss without additional pass through model
 """
 function validation_loss(SkipGAN::SkipGANomaly, real_input; weights=[1,50,1])
+	testmode!(SkipGAN)
```
vitskvara: Again

vitskvara: And for all the other functions the same.

src/models/adVAE.jl
```diff
@@ -0,0 +1,187 @@
...
+	# training info logger
+	history = MVHistory()
+	# prepare for early stopping
+	best_adVAE = deepcopy(advae)
+	patience = params.patience
+	best_val_loss = 1e10
+	val_batches = length(val_loader)
+
+	# ADAMW(η = 0.001, β = (0.9, 0.999), decay = 0) = Optimiser(ADAM(η, β), WeightDecay(decay))
+	opt = haskey(params, :decay) ? ADAMW(params.lr, (0.9, 0.999), params.decay) : ADAM(params.lr)
```
vitskvara: You need two optimisers, each for the separate set of parameters.

vitskvara: See the implementation of AAE.

src/models/adVAE.jl
```diff
@@ -0,0 +1,187 @@
...
+	ps_step2 = Flux.params(advae.encoder)
+
+
+	progress = Progress(length(train_loader))
+	for (iter, X) in enumerate(train_loader)
+		# training step 1
+		loss_1, back = Flux.pullback(ps_step1) do
+            loss(advae|>gpu , getobs(X)|>gpu, γ=params.gamma, λ=params.lambda, mx=params.mx, mz=params.mz)
+		end
+		grad_step1 = back((1f0, 0f0, 0f0, 0f0))
```
vitskvara, line 143: Why do you do this?

src/models/adVAE.jl
```diff
@@ -0,0 +1,187 @@
...
+
+	ps_step1 = Flux.params(advae.generator, advae.transformer)
+	ps_step2 = Flux.params(advae.encoder)
+
+
+	progress = Progress(length(train_loader))
+	for (iter, X) in enumerate(train_loader)
+		# training step 1
+		loss_1, back = Flux.pullback(ps_step1) do
+            loss(advae|>gpu , getobs(X)|>gpu, γ=params.gamma, λ=params.lambda, mx=params.mx, mz=params.mz)
```
vitskvara: What if you train on tabular data? Than this does not work. Move the model to gpu outside the fit function. Also, I would create a separate loss for tabular and image data, where the other is going to move them on gpu, and set it as an argument of the fit! function. See the vae.jl run scripts for tabular and image data.

src/models/adVAE.jl
```diff
@@ -0,0 +1,187 @@
...
+					global iters = iter - params.check_every*params.patience
+					break
+				end
+			end
+		end
+		if iter == params.iters
+			global iters = params.iters
+		end
+	end
+	return history, best_adVAE, sum(map(p->length(p), ps_step1))+sum(map(p->length(p), ps_step2)), iters
```
vitskvara: If you have defined the model correctly with functors etc, you should only need to call `sum(map(p->length(p), Flux.params(best_adVAE)))`

src/models/adVAE.jl
```diff
@@ -0,0 +1,187 @@
...
+function (advae::adVAE)(x)
+    μ, Σ = advae.encoder(x)
+    z = μ + Σ * randn(Float32)
+    xᵣ = advae.generator(z)
+end
+
+kl_divergence(μ, Σ) = 0.5 * sum(1 .+ log.(Σ.^2) - μ.^2  - Σ.^2) # actually it is -kl_divergence
+kl_divergence(μ₁, Σ₁, μ₂, Σ₂) = sum(log.(Σ₂ ./ Σ₁) + (Σ₁.^2 + (μ₁ - μ₂).^2) ./ (2*Σ₂.^2) .- 0.5)
+
+function loss(advae::adVAE, x; γ=1e-3, λ=1e-2, mx=1, mz=1)
```
vitskvara, line 69: Maybe, if this does not work, define separate loss functions for 1st and 2nd optimization step?

src/models/adVAE.jl
```diff
@@ -0,0 +1,187 @@
...
+    μₜᵣ, Σₜᵣ = advae.encoder(xₜᵣ)
+
+    # 𝓛 for generator => 𝓛 = 𝓛_z + 𝓛_zₜ 
+    𝓛_z = Flux.Losses.mse(x, xᵣ) .+ γ * kl_divergence(μᵣ, Σᵣ)
+    𝓛_zₜ = max(0, (mx - Flux.Losses.mse(xᵣ, xₜᵣ))) + γ * max(0, (mz - kl_divergence(μₜᵣ, Σₜᵣ)))
+    𝓛 = 𝓛_z + 𝓛_zₜ
+    # 𝓛ₜ for transformer 
+    𝓛ₜ = kl_divergence(μ, Σ, μₜ, Σₜ)
+    # 𝓛ₑ for encoder
+    𝓛ₑ = Flux.Losses.mse(x, xᵣ) .+ γ * kl_divergence(μ, Σ)
```
vitskvara: Are you sure that the +- signs are all correct and every loss is written in such for that it is minimized? Because here it seems that the kl_divergence is actually maximized when it is with a - sign.

vitskvara (changes requested):

scripts/experiments_images/adVAE.jl
TheMrGhostman: yes, my bad

src/models/GANomaly.jl
TheMrGhostman: ok, i will change it

src/models/adVAE.jl
TheMrGhostman: you mean this `grad_step1 = back((1f0, 0f0, 0f0, 0f0))` ?

it helps me trace individual losses. I returns all those losses but i am backpropagating only first one.

src/models/adVAE.jl
TheMrGhostman: it works on tabular data. There is no problem. 
Moreover when you dont have gpu on your system, it automaticly works only on cpu with no error or warning.
But if you want i will chagne it.

src/models/adVAE.jl
TheMrGhostman: changed

src/models/adVAE.jl
TheMrGhostman: i will try

src/models/adVAE.jl
TheMrGhostman: yes, in original paper there is mistake in +/- but i corrected it. mybe i should call `𝓛_kld instead of kl_divergence` because correctrly it is `-kld` 
-> `kl_divergence(μ, Σ) = 0.5 * sum(1 .+ log.(Σ.^2) - μ.^2  - Σ.^2) # actually it is -kl_divergence`

src/models/adVAE.jl
TheMrGhostman: I think that Tomáš told me that when you have same hyperparameters (like lr and betas) you can use only one optimizer for both models.

vitskvara (approved):

janfrancu:
Please bump pkg version to `0.3.0`.

PR #47, opened 2020-10-22
Create ocsvm.jl

TheMrGhostman:
creating ocsvm ensemble for image data.


scripts/experiments_images/ocsvm.jl
```diff
@@ -0,0 +1,112 @@
...
+		default = 10
+		help = "number of anomaly classes"
+end
+parsed_args = parse_args(ARGS, s)
+@unpack dataset, max_seed, anomaly_classes = parsed_args
+
+modelname = "OCSVM"
+
+function sample_params()
+	par_vec = (round.([10^x for x in -4:0.1:2],digits=5),["poly", "rbf", "sigmoid"])
```
vitskvara: Add the `nu` parameter with values `[0.01, 0.5, 0.99]`

scripts/experiments_images/ocsvm.jl
```diff
@@ -0,0 +1,112 @@
...
+    return Flux.mean(joint_pred, dims=1)
+end
+
+function fit(data, parameters, n_models=10)
+    # construct model - constructor should only accept kwargs
+    models = [GenerativeAD.Models.OCSVM(;parameters...) for _ = 1:n_models]
+	
+	# sumbsample and fit train data
+	tr_data = data[1][1]
+    M,N = size(tr_data)
```
vitskvara: This is not gonna work, image data have size = (w,h,c,n).

scripts/experiments_images/ocsvm.jl
```diff
@@ -0,0 +1,112 @@
...
+    for (i,model) in enumerate(models)
+        joint_pred[i,:] = predict(model, data)
+    end
+    return Flux.mean(joint_pred, dims=1)
+end
+
+function fit(data, parameters, n_models=10)
+    # construct model - constructor should only accept kwargs
+    models = [GenerativeAD.Models.OCSVM(;parameters...) for _ = 1:n_models]
+
```
vitskvara, line 56: What I would do is that I would reshape all data here, e.g. `data[1][1] = reshape(data[1][1], :, size(data[1][1],4))` (the same for (val and tst data).

vitskvara: Then the rest is gonna work as you have it written already.

scripts/experiments_images/ocsvm.jl
```diff
@@ -0,0 +1,112 @@
...
+while try_counter < max_tries
+	parameters = sample_params()
+
+	for seed in 1:max_seed
+		for i in 1:anomaly_classes
+			savepath = datadir("experiments/images/$(modelname)/$(dataset)/ac=$(i)/seed=$(seed)")
+
+			data = GenerativeAD.load_data(dataset, seed=seed, anomaly_class_ind=i)
+
```
vitskvara: Ignore my second comment, it is actually better to reshape the data here, e.g. write a nice function for it.

scripts/experiments_images/ocsvm.jl
```diff
@@ -0,0 +1,112 @@
...
+	end
+
+	# construct return information - put e.g. the model structure here for generative models
+	training_info = (
+		fit_t = fit_t,
+		model = nothing
+		)
+
+	# now return the different scoring functions
+	training_info, [(x->joint_predictiion(models, x), parameters)]
```
vitskvara: typo

scripts/experiments_images/ocsvm.jl
```diff
@@ -0,0 +1,112 @@
...
+    info = []
+    for (model, data) in zip(models, data_splits)
+        push!(info, fit!(model, data))
+    end
+    return info
+end
+
+function joint_prediction(models, data)
+    joint_pred = Array{Float32}(undef, length(models), size(data,2))
+    for (i,model) in enumerate(models)
```
vitskvara: Is it not more economical to do the prediction here sample-wise? That way you only need to allocate a vector as you can already do the average in each step.

vitskvara (changes requested):
Also add `ocsvm_run.sh` file

scripts/experiments_images/ocsvm.jl
TheMrGhostman: it will work because i reshaped images before

scripts/experiments_images/ocsvm.jl
TheMrGhostman: fixed

scripts/experiments_images/ocsvm.jl
TheMrGhostman: i added function to utils and reshaping will take place here

vitskvara (approved):

vitskvara:
Added ocsvm, knn and abod for image data.

PR #50, opened 2020-10-23
JacoDeco alternative score 

vitskvara:


vitskvara:
I found a bug in savenames of all generative models where they are different for model files and for score files. I will need to fix this first since it may break a lot of things including selection of best models for ensembles/two stage models.

PR #55, opened 2020-10-29
AdVAE + 2 stage

vitskvara:


scripts/experiments_images/2stage-vae-knn.jl
```diff
@@ -0,0 +1,127 @@
...
+        arg_type = Int 
+        default = 1
+        help = "index of model in table vae-table.csv"
+end
+parsed_args = parse_args(ARGS, s)
+@unpack dataset, max_seed, anomaly_classes, model_index = parsed_args
+
+#######################################################################################
+################ THIS PART IS TO BE PROVIDED FOR EACH MODEL SEPARATELY ################
+modelname = "vae+knn"
```
vitskvara: change to vae-knn, this is ugly

scripts/experiments_images/2stage-vae-knn.jl
```diff
@@ -0,0 +1,127 @@
+using DrWatson
```
vitskvara: I would change this to a universal 2stage-knn script and add the file with the list of models as an argument. Then we can run it for universal encoding model.

vitskvara: Of course the savepath is gonna unique for each combination of encoding model and knn, probably deciphered from the model list filename.

scripts/experiments_images/2stage-vae-ocsvm.jl
```diff
@@ -0,0 +1,118 @@
+using DrWatson
```
vitskvara: Here the same, make it universal for different encoding models.

scripts/experiments_images/2stage-vae-knn.jl
vitskvara: Actually, this is gonna be detected automatically from the model list filename.

scripts/vae_for_two_stage.jl
```diff
@@ -0,0 +1,104 @@
+using DrWatson
+@quickactivate
+using BSON
+using DataFrames
+using ValueHistories
+using Statistics
+using CSV
+
+function fix_info_name(name)
```
vitskvara: this is not gonna be needed soon.

scripts/vae_for_two_stage.jl
```diff
@@ -0,0 +1,104 @@
...
+				i += 1
+			end
+		end
+	end
+	return df
+end
+
+
+path = "/home/skvarvit/generativead/GenerativeAD.jl/data/experiments/images/vae/"
+models = models_and_params(path)
```
vitskvara: This seems overly complicated, you could do just 
```julia
files = GenerativeAD.Evaluation.collect_files(path)
models = filter(f->occursin("model", f), files)
```

vitskvara: And then get the paths, seeds, dataset names etc from the strings.

scripts/vae_for_two_stage.jl
```diff
@@ -0,0 +1,104 @@
...
+	models = []
+	
+	for (root, dirs, files) in walkdir(path_to_model)
+		for file in files
+			push!(directories, root)
+		end
+	end
+	
+	for dir in unique(directories)
+		fs = filter(x->(startswith(x, "model")), readdir(dir))
```
vitskvara: You could probably just map them using the seed, but now all modelfiles should contain the requested fields.

scripts/vae_for_two_stage.jl
```diff
@@ -0,0 +1,104 @@
...
+			seed = Int64[], 
+			loss_val= Float32[]
+		)
+	else
+		df = DataFrame(
+			path = String[], #path to model / encodings
+			params = String[],
+			dataset = String[], 
+			seed = Int64[], 
+			loss_val= Float32[]
```
vitskvara: I woul go from `loss_val` to `criterion_val`, so that we can use different sorting criterions (e.g. this, of validation auc etc.)

src/models/utils/utils.jl
```diff
@@ -105,4 +108,59 @@ function clip_weights!(ps::Flux.Zygote.Params,low::Real,high::Real)
...
+
+
+function load_encoding(model="vae_img", data; dataset::String="MNIST", anomaly_class::Int=1, seed::Int=1, model_index::Int=1)
+    # load csv
+	df = CSV.read(datadir("$(model)_tab.csv")) 
+	df = return_best_n(df, 10, dataset)
+    # checking if model configuration was trained on all classes and all seeds
+    n_comb = 500
+    check_comb = sum([sum(df.ind .== i) for i=1:10])
+    if check_comb < n_comb
```
vitskvara: I would not worry about this, for a given dataset/seed/class combination, we just take the best n, no matter what.

src/models/utils/utils.jl
```diff
@@ -105,4 +118,60 @@ function clip_weights!(ps::Flux.Zygote.Params,low::Real,high::Real)
...
+    for i=1:10
+        tmp = df[df.params .== top.params[i],:]
+        ind = i.*ones(size(tmp,1))
+        push!(df_top, hcat(tmp, DataFrame(ind=ind)))
+    end
+    return vcat(df_top...)
+end
+
+
+function load_encoding(model="vae_tabular", data; dataset::String="iris", seed::Int=1, model_index::Int=1)
```
vitskvara: I would change this so that the input argument is the actual table with model performances

scripts/experiments_images/2stage_knn.jl
```diff
@@ -0,0 +1,133 @@
...
+@unpack dataset, max_seed, tab_name, anomaly_classes = parsed_args
+
+#######################################################################################
+################ THIS PART IS TO BE PROVIDED FOR EACH MODEL SEPARATELY ################
+
+sp = split(tab_name, "_")
+enc = sp[1]
+criterion = lowercase(sp[2])
+
+modelname = "$(enc)+knn"
```
vitskvara: I still dont like the plus, I would go with underscore or -.

scripts/experiments_images/2stage_knn.jl
```diff
@@ -0,0 +1,133 @@
...
+
+                data = GenerativeAD.load_data(dataset, seed=seed, anomaly_class_ind=i)
+                data, encoding_name = GenerativeAD.Models.load_encoding(tab_name, data, dataset=dataset, anomaly_class=i, seed=seed, model_index=mi)
+
+                # here, check if a model with the same parameters was already tested
+                @info "Trying to fit $modelname on $dataset with parameters $(parameters)..."
+                if GenerativeAD.check_params(savepath, merge(parameters, aux_info))
+                    training_info, results = fit(data, parameters, aux_info)
+                    # here define what additional info should be saved together with parameters, scores, labels and predict times
+                    save_entries = merge(training_info, (modelname = modelname,
```
vitskvara: Save the full path to the original modelfile and also the original model parameters.

vitskvara: Because the input table might change and then the index is useless.

scripts/experiments_images/2stage_ocsvm.jl
```diff
@@ -0,0 +1,141 @@
...
+@unpack dataset, max_seed, tab_name, anomaly_classes = parsed_args
+
+#######################################################################################
+################ THIS PART IS TO BE PROVIDED FOR EACH MODEL SEPARATELY ################
+
+sp = split(tab_name, "_")
+enc = sp[1]
+criterion = lowercase(sp[2])
+
+modelname = "$(enc)+ocsvm"
```
vitskvara: enc_ocsvm

scripts/experiments_tabular/adVAE.jl
```diff
@@ -85,7 +85,7 @@ function fit(data, parameters)
 		)
 
 	return training_info, 
-	[(x -> GenerativeAD.Models.anomaly_score(advae|>cpu, x; dims=1, L=100), merge(parameters, (L=100, )))]
+	[(x -> GenerativeAD.Models.anomaly_score(advae|>cpu, x; dims=1, L=100)[:], merge(parameters, (L=100, )))]
```
vitskvara, line 88: Or you could use vec(...) instead of [:]

src/models/Models.jl
```diff
@@ -12,6 +12,7 @@ using MLDataPattern
 using ProgressMeter: Progress, next!
 
 include("utils/utils.jl")
+include("utils/2stage.jl")
```
vitskvara: I would probably like it more if it was two_stage instead of 2stage, files beginning with numbers don't look nice.

vitskvara:
This might be the parallel script:
```
#!/bin/bash
# This runs parallel experiments over all datasets.
# USAGE EXAMPLE
# 	./run_two_stage_parallel.sh knn vae_LOSS_images 3 1 2 datasets_tabular.txt
# Run from this folder only.
MODEL=$1                     # second stage model
TABLE_NAME=$2          # name of table with stored results
NUM_SAMPLES=$3	# how many repetitions
MAX_SEED=$4		# how many folds over dataset
NUM_CONC=$5		# number of concurrent tasks in the array job
DATASET_FILE=$6	# file with dataset list

LOG_DIR="${HOME}/logs/two_stage_${MODEL}"

if [ ! -d "$LOG_DIR" ]; then
	mkdir $LOG_DIR
fi

while read d; do
	# submit to slurm
    sbatch \
    --array=1-${NUM_SAMPLES}%${NUM_CONC} \
    --output="${LOG_DIR}/${d}-%A_%a.out" \
     ./2stage_${MODEL}_run.sh $MAX_SEED $d ${TABLE_NAME} 10
done < ${DATASET_FILE}

```

vitskvara:
Don't forget to make the run scripts executable (chmod +x runscript.sh)

scripts/experiments_images/2stage_ocsvm.jl
TheMrGhostman: ok, changed

scripts/experiments_images/2stage_knn.jl
TheMrGhostman: done

scripts/vae_for_two_stage.jl
TheMrGhostman: true, it would be nicer but that script will run only once. So i hope it won't be issue

src/models/Models.jl
TheMrGhostman: done

scripts/experiments_images/2stage_knn.jl
TheMrGhostman: i thought this is enough 

```
save_entries = merge(training_info, (modelname = modelname, 
				     seed = seed, 
				     dataset = dataset, 
				     anomaly_class = i, 
				     encoding_name=encoding_name,
				     model_index=mi,
				     criterion=criterion))
```
those informations contain all you need, like trace to original VAE-type model used for encoding
-> `modelname[1]/dataset/seed=seed/ac=anomaly_class/encoding_name = vae/MNIST/seed=1/ac=1/model_***.bson`

vitskvara: So that is saved in `encoding_name`? That is a confusing name. Also, if you save the original model parameters here, you won't have to retrieve them from the original file later, which might be really time-consuming since you have to load the whole bson with the neural net object and load all the dependencies.

PR #56, opened 2020-10-29
Eval extension & fixes

janfrancu:
- [x] adding argument for `run_eval` script

- [x] adding std over top 10  hyperparameters

- [x] adding threshold on number of seed/anomaly_class samples per given hyperparameter

- [x] fix: undersampling fails on image data due to `ocsvm` having the same name as for tabular data

- [x] add arg modelkey to group on - allow for grouping models into groups such as classical, flows, autoencoders, ... **is not part of this pull request as it can be done with the current code, just not as nicely**

### DIFF
--- a/src/evaluation/show.jl
+++ b/src/evaluation/show.jl
@@ -14,7 +14,12 @@ Last row of the dataframe contains average rank of each model.
 function rank_table(df::DataFrame, metric_col=:tst_auc)
 	# check if column names are present
 	(!(String(metric_col) in names(df)) || !("modelname" in names(df)) || !("dataset" in names(df))) && error("Incorrect column names.")
-	(!(String(metric_col)*"_std" in names(df))) && error("DataFrame does not contain std for the given metric.")
+	
+	std_column = String(metric_col)*"_std"
+	(!(std_column in names(df))) && error("DataFrame does not contain std for the given metric.")
+	
+	top_10_std_column = String(metric_col)*"_top_10_std"
+	(!(top_10_std_column in names(df))) && error("DataFrame does not contain std for the given metric.")
 
 	# get all the models that are present in the dataframe
 	all_models = unique(df.modelname)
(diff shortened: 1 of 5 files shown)


src/evaluation/show.jl
```diff
@@ -39,22 +44,27 @@ function rank_table(df::DataFrame, metric_col=:tst_auc)
...
+	push!(rt, ["MEAN_TOP10_STD", top_10_std...])
+	
+	# round all values in the table before computing rank
+	rt[:, 2:end] .= round.(rt[:, 2:end], digits=2)
+	
 	# add average rank
 	mask_nan_max = (x) -> (isnan(x) ? -Inf : x)
 	rs = zeros(size(rt, 2) - 1)
-	for row in eachrow(rt)
+	for row in eachrow(rt[1:end-2,:])
```
janfrancu, line 63: This formerly affected the final ranking, as it took into account the `std` rows as another dataset. It would be better if in the future these two dataframes were separated.

src/evaluation/show.jl
```diff
@@ -39,22 +44,27 @@ function rank_table(df::DataFrame, metric_col=:tst_auc)
...
+		filter(x -> (x.modelname == m), df)[std_column]) for m in names(rt[:,2:end])]
 	push!(rt, ["MEAN_STD", mean_std...])
 
+	# add top 10 std (more precisely mean over dataset)
+	top_10_std = [mean(
+		filter(x -> (x.modelname == m), df)[top_10_std_column]) for m in names(rt[:,2:end])]
+	push!(rt, ["MEAN_TOP10_STD", top_10_std...])
+	
+	# round all values in the table before computing rank
+	rt[:, 2:end] .= round.(rt[:, 2:end], digits=2)
```
janfrancu, line 58: Mean values of standard deviations are computed in full precision now and this rounds all the values before ranks computation.

PR #57, opened 2020-11-02
Ensemble computation.

janfrancu:
- [x] initial prototype

- [x] support both image and tabular datasets

- [x] load evaluation files rather than original files

- [x] figure out where to store the results -  ~~experiments~~ or separate folder

- [x] run scripts


vitskvara:
I would store the results to a dir parallel to images/tabular, e.g. images_ensembles/tabular_ensembles.

janfrancu:
This is as far as we will explore ensembles for now. Closing.

PR #58, opened 2020-11-03
Add max_mean aggregation to evaluation.

janfrancu:
- [x] initial draft

- [x] check if output is correct

- [x] ~~criteria on number of samples/repetition, i.e. implement `min_samples`~~ does not make much sense in max-mean aggr

- [x] ~~modify `evaluate_performance` script to take this new aggregation into account~~ exposed only the default mean-max method

### DIFF
--- a/src/evaluation/stats.jl
+++ b/src/evaluation/stats.jl
@@ -118,6 +118,8 @@ columns of different types
 	+ `dsamples_valid` - number of sampled hyperparameters with enough runs
 When nonempty `downsample` dictionary is specified, the entries of`("model" => #samples)`, specify
 how many samples should be taken into acount. These are selected randomly with fixed seed.
+Optional arg `min_samples` specifies how many seed/anomaly_class combinations should be present
+in order for the hyperparameter's results be considered statistically significant.
 """
 function aggregate_stats(df::DataFrame, criterion_col=:val_auc; 
 							min_samples=("anomaly_class" in names(df)) ? 30 : 3, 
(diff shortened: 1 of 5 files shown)


src/evaluation/stats.jl
```diff
@@ -170,3 +172,70 @@ function aggregate_stats(df::DataFrame, criterion_col=:val_auc;
 	end
 	vcat(results...)
 end
+
+
+"""
+aggregate_stats_max_mean(df::DataFrame, criterion_col=:val_auc; undersample=Dict("ocsvm" => 100))
```
vitskvara: Change the docstring to downsample (in the original aggregate stats its missing as well).

src/evaluation/stats.jl
```diff
@@ -170,3 +172,70 @@ function aggregate_stats(df::DataFrame, criterion_col=:val_auc;
...
+
+Chooses the best hyperparameters for each seed/anomaly_class combination by `criterion_col`
+and then aggregates the metrics over seed/anomaly_class to get the final results. The output 
+is a DataFrame of maximum #datasets*#models rows with
+columns of different types
+- identifiers - `dataset`, `modelname`, `phash`, `parameters`
```
vitskvara: When i tried to run this, I did not get the columns with phash and parameters.

vitskvara: Which makes sense, since they are now not universal for all selected models. But you should edit the docstring.

vitskvara:
So I wrote the following test where I recompute the numbers in a different, slightly more transparent but uglier way. The numbers fit exactly for most models and are slightly off for "knn" or "vae", which I think is caused by more than one seed for theses two models. Correct me if I am wrong. From this, I would approve the pull request.

```julia
using DrWatson
@quickactivate
using CSV
using BSON
using Random
using FileIO
using DataFrames
using Statistics

using GenerativeAD.Evaluation: _prefix_symbol, PAT_METRICS, aggregate_stats
using GenerativeAD.Evaluation: rank_table, print_rank_table, aggregate_stats_max_mean

filename = "/home/skvarvit/generativead/GenerativeAD.jl/data/evaluation/images_eval.bson"
criterion_metric =  "val_auc"
downsample=Dict()
df = load(filename)[:df]
df_agg = aggregate_stats_max_mean(df, Symbol(criterion_metric), downsample=downsample)

# test if the numbers fit
dataset = "CIFAR10"
modelname = "knn"
subdf = filter(r->r[:dataset]==dataset && r[:modelname]==modelname, df)
smeasure = "val_auc"
pmeasure = "tst_auc"
res = []
for ac in 1:10
	for seed in 1:1
		ssdf = sort(filter(r->r[:seed] == seed && r[:anomaly_class]==ac, subdf), order(smeasure, rev=true))
		push!(res, ssdf[1, pmeasure])
	end
end
orig_df = filter(r->r[:dataset]==dataset && r[:modelname]==modelname, df_agg_mm)[!, vcat([50, 51], 1:49)]

# these two numbers should be the same
mean(res)
orig_df[pmeasure]
```

vitskvara:
Actually by adding all 5 seeds and a check for empty dataframe I now conclude that the reported numbers are correct.

src/evaluation/stats.jl
```diff
@@ -170,3 +172,70 @@ function aggregate_stats(df::DataFrame, criterion_col=:val_auc;
...
+agg_cols = vcat(_prefix_symbol.("val", BASE_METRICS), _prefix_symbol.("tst", BASE_METRICS))
+agg_cols = vcat(agg_cols, _prefix_symbol.("val", PAT_METRICS), _prefix_symbol.("tst", PAT_METRICS))
+top10_std_cols = _prefix_symbol.(agg_cols, "top_10_std")
+
+agg_keys = ("anomaly_class" in names(df)) ? [:seed, :anomaly_class] : [:seed]
+# choose best for each seed/anomaly_class cobination and average over them
+results = []
+for (dkey, dg) in pairs(groupby(df, :dataset))
+    for (mkey, mg) in pairs(groupby(dg, :modelname))
+        partial_results = []
```
vitskvara: If we have agreed on seed 1 only, you should filter here somehow.

vitskvara (approved):

src/evaluation/stats.jl
janfrancu: I hesitate to put such filter into the aggregation itself and leave this to the user to filter beforehand.

src/evaluation/stats.jl
janfrancu: Updated.

janfrancu:
Thanks for the feedback. Merging.

PR #76, opened 2020-12-10
Leave one in

vitskvara:
What needs to be cleaned here?


TheMrGhostman:
i am testing functionality with loo, to be sure we don't break something

TheMrGhostman:
This should be merged before F-AnoGAN-GP. That branch is more updated, so we can avoid another conflict

PR #85, opened 2021-03-09
MNIST-C & MVTec-AD

vitskvara:
I will also add MvTec-AD here.

MNIST-C consists of 15 corruption categories that change the original MNIST dataset. This basically presents us with 15 new datasets, where we will detect non-corrupted vs corrupted.

 "brightness"
 "canny_edges"
 "dotted_line"
 "fog"
 "glass_blur"
 "impulse_noise"
 "motion_blur"
 "rotate"
 "scale"
 "shear"
 "shot_noise"
 "spatter"
 "stripe"
 "translate"
 "zigzag"


vitskvara:
MVTec-AD consists of 15 subproblems. Each has a few (10s~) normal samples. Each subproblem has a set of different kinds of anomalies, e.g. in `bottle`, there are `broken_large,  broken_small,  contamination` kinds of anomalies. The images are large - 900x900x3, but there is not a lot of them.

bottle 
capsule
grid      
leather      
metal_nut 
tile        
transistor  
zipper
cable   
carpet   
hazelnut 
pill       
screw       
toothbrush  
wood

vitskvara:
Currently the else-if in `load_data` is really ugly, but it seems to be the shortest way to do it.

vitskvara:
VAE on MNIST-C seems to be working.

vitskvara:
But I am getting OOM errors for MVTec (on the RCI cluster).

vitskvara:
I am currently using a Dense layer on the output of decoder that comes after the convolutions and which I use to compute the vectorized mean and scalar variance of the output. However, this now tries to use a W matrix with one dimension of 3x900^2, which already fails at initialization with Glorot. So I will rewrite this piece without the last Dense layer, probably estimating the scalar variance from an additional conv channel.

scripts/experiments_images/run_parallel_no_class.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# This runs parallel experiments over all datasets.
+# USAGE EXAMPLE
+# 	./run_parallel.sh pidforest 3 1 2 datasets_tabular.txt 0.01
```
janfrancu: Please update the usage example.

src/datasets/basics.jl
```diff
@@ -93,8 +93,20 @@ function load_data(dataset::String, ratios=(0.6,0.2,0.2); seed=nothing,
...
-	elseif dataset in other_datasets
+	elseif dataset in other_datasets # other tabular datasets
 		data_normal, data_anomalous = load_other_data(dataset; standardize=true, kwargs...)
+	elseif dataset =="MNIST-C"
+		data_normal, data_anomalous = load_mnist_c_data(; kwargs...)
+	elseif dataset == "MVTec-AD"
+		data_normal, data_anomalous = load_mvtec_ad_data(; kwargs...)
+	elseif occursin("MNIST-C", dataset)
+		sd = split(dataset, "_")
+		category = (length(sd) >= 2) ? dataset[9:end] : "brightness"
```
janfrancu: Add comment as to why this case has to be handled separately.

janfrancu: Correct me if I am wrong, but looking at the possibilities in `scripts/experiments_images/datasets_mnistc.txt` this never chooses `MNIST-C_fog` or others without underscore.

src/datasets/basics.jl
```diff
@@ -93,8 +93,20 @@ function load_data(dataset::String, ratios=(0.6,0.2,0.2); seed=nothing,
...
+		data_normal, data_anomalous = load_mnist_c_data(; category=category, kwargs...)
+	elseif occursin("MVTec-AD", dataset) 
+		sd = split(dataset, "_")
+		category = (length(sd) >= 2) ? dataset[10:end] : "bottle"
```
janfrancu: Not sure about this either.

src/datasets/img.jl
```diff
@@ -28,3 +28,79 @@ function load_mldatasets_data(dataset::String; anomaly_class_ind::Int=1)
...
+
+"""
+	load_images_rgb(path)
+
+Load all images in a path.
+"""
+function load_images_rgb(path)
+	imgs = []
+	for f in readdir(path, join=true)
+		try # some files are corrupted
```
janfrancu, line 98: Has this already been reported somewhere or are we the first one to encounter it?

janfrancu (changes requested):

src/datasets/basics.jl
vitskvara:
```julia
dataset = "MNIST-C_fog"
sd = split(dataset, "_")
category = (length(sd) >= 2) ? dataset[9:end] : "brightness"
```
`category` now evaluates to `"fog"`

scripts/experiments_images/run_parallel_no_class.sh
vitskvara: Done

src/datasets/img.jl
vitskvara: Ok, so this was a problem with my partial download of the dataset. Should I remove the try catch block?

janfrancu (approved):

PR #86, opened 2021-03-17
Theoretical POC of Bayesian optimization using skopt

### LINKED ISSUE #82: Bayesian optimization
Start with tabular images and some simple models. 

- [ ] Update the run scripts so inputs are params.
- [ ] Put the sampling functions in one place.
- [ ] Use 50 already trained datapoints for initialization and use it to compute another 50 samples of the hyperparam space.
- [ ] Use AUC as criterion.

janfrancu:
There are some naming conventions, that have to be followed, but in the end it does not look that atrocious.
Let's hope it works.
- [x] Concept
- [x] Make it work on at least one model.


janfrancu:
Following points are yet to be addressed
- [x] better script for Bayesian cache creation (automate over datasets)
- [x] change interface in runscripts from bayes = [true|false] to sampling = [bayes|random]
- [x] location of the newly trained models (separate experiment folder?)
- [x] How to treat multiple scores? Max aggregation or just choose one? This has to be resolved during cache creation.
- [ ] FIX: `edit_params` will modify params after they were written to the cache. -> won't fix
- [x] aggregation of metrics on image datasets should reflect the evaluation protocol -> configurable with `run_agg` argument in `objective_value`

janfrancu:
Models tested so far
- tabular data
    - [x] OCSVM - initial experiments showed overall improvement
    - [x] MAF - initial experiments showed overall improvement
    - [x] RealNVP
    - [x] SPTN - limited number of parameters to be tuned, some of them may just fail due to time limits set in place
    - [x] PIDforest - missing `edit_params` in Bayesian optimization leads to fewer trained models, the search space is thus really reduced on some datasets 
    - [x] VAE - done
    - [x] WAE - done
    - [x] AAE - done
    - [ ] fmGAN - caches built, running first experiments
    - [ ] GAN - caches built
    - [x] GANomaly
    - [x] adVAE
    - [ ] HBOS - caches built
    - [ ] IF - caches built
    - [ ] LOF - caches built
    - [ ] knn - caches built
    - [ ] LODA - caches built
    - [ ] REPEN
    - [ ] DAGMM
    - [ ] ABOD - caches built
    - [ ] MOGAAL
    - [ ] ?OCSVM_RBF - there could be some improvement from increased range on `nu` parameters
    - [ ] ?2STAGE
- image data
    - [ ] ocsvm - tested creation of cache only

scripts/bayes/bayes_cache.jl
```diff
@@ -0,0 +1,201 @@
+"""
+Example script how to manually create Bayesian cache from already computed results. 
+Intended to run interactively as user input is needed to rewrite parts of it.
+Workflow:
+    1. pick a model and dataset
```
vitskvara, line 7: I would add 
```
0. Install `scikit-optimize`.
```
:-)

scripts/experiments_tabular/maf.jl
```diff
@@ -15,13 +17,17 @@ s = ArgParseSettings()
...
 		arg_type = String
 		help = "dataset"
-    "contamination"
-    	arg_type = Float64
-    	help = "contamination rate of training data"
-    	default = 0.0
+	"sampling"
+		default = "random"
+		arg_type = String
+		help = "sampling of hyperparameters"
```
vitskvara, line 23: Maybe add the options, e.g. "random"/"bayes"

scripts/experiments_tabular/maf.jl
```diff
@@ -70,16 +96,26 @@ end
 try_counter = 0
 max_tries = 10*max_seed
 cont_string = (contamination == 0.0) ? "" : "_contamination-$contamination"
+sampling_string = sampling == "bayes" ? "_bayes" : ""
```
vitskvara, line 99: Seriously, we should put this together as one function in GenerativeAD.jl, I don't see much variance of this block of code between my models and the others I think are more or less the same.

scripts/experiments_tabular/pidforest_run.sh
```diff
@@ -1,21 +1,21 @@
...
+# load virtualenv containing skopt+sklearn+pidforest
+source ${HOME}/sklearn-env/bin/activate
+export PYTHON="${HOME}/sklearn-env/bin/python"
 
-# PyCall needs to be rebuilt if environment changed
-julia --project -e 'using Pkg; Pkg.build("PyCall"); @info("SETUP DONE")'
+julia --project -e 'using Pkg; Pkg.instantiate();'
 
-julia --project ./pidforest.jl $MAX_SEED $DATASET $CONTAMINATION
+julia --project ./pidforest.jl $MAX_SEED $DATASET $HP_SAMPLING $CONTAMINATION
```
vitskvara, line 24: For me, this does not work unless `HP_SAMPLING` and `MAX_SEED` is in brackets - e.g. `${MAX_SEED}`

scripts/experiments_tabular/wae_full.jl
```diff
@@ -130,8 +155,18 @@ if abspath(PROGRAM_FILE) == @__FILE__
...
+			parameters = GenerativeAD.bayes_params(
+									create_space(), 
+									dataset_folder,
+									sample_params; add_model_seed=true)
+		else
+			parameters = sample_params()
+		end
 
 	    for seed in 1:max_seed
 			savepath = datadir("experiments/tabular$cont_string/$(modelname)/$(dataset)/seed=$(seed)")
```
janfrancu: Change this to `savepath = joinpath(dataset_folder, "seed=$(seed)")`. Currently the models are stored under `experiments` instead of `experiments_bayes`.

scripts/experiments_tabular/aae_full.jl
```diff
@@ -122,8 +144,18 @@ if abspath(PROGRAM_FILE) == @__FILE__
...
+			parameters = GenerativeAD.bayes_params(
+									create_space(), 
+									dataset_folder,
+									sample_params; add_model_seed=true)
+		else
+			parameters = sample_params()
+		end
 
 	    for seed in 1:max_seed
 			savepath = datadir("experiments/tabular$cont_string/$(modelname)/$(dataset)/seed=$(seed)")
```
janfrancu: Change this to savepath = joinpath(dataset_folder, "seed=$(seed)"). Currently the models are stored under experiments instead of experiments_bayes.

scripts/experiments_tabular/wae_full.jl
vitskvara: Good catch, fortunately the cache is still updated, so I just had to move the models together with the cache.

PR #87, opened 2021-03-23
Autoencoders

TheMrGhostman:
tested on rci


scripts/experiments_images/DeepSVDD.jl
```diff
@@ -87,7 +87,7 @@ function fit(data, parameters)
 			ae_iters=10000
 		)
 	)
-	svdd = GenerativeAD.Models.conv_ae_constructor(;all_parameters...) |> gpu
```
vitskvara, line 90: Why was an autoencoder constructor used in the first place?

src/models/ae.jl
```diff
@@ -0,0 +1,160 @@
...
+	decoder = build_mlp(zdim, hdim, idim, nlayers, activation=activation, lastlayer="linear")
+
+	(init_seed !== nothing) ? Random.seed!() : nothing
+
+	return AE(encoder, decoder)
+end
+
+function anomaly_score(ae::AE, real_input; dims=(1,2,3), to_testmode::Bool=true)
+	(to_testmode == true) ? Flux.testmode!(ae) : nothing
+	score = vec(Flux.sum((ae(real_input) .- real_input).^2, dims=dims))
```
vitskvara: I would probably average the sum of squares so that different datasets with different dimensions are comparable.

src/models/ae.jl
```diff
@@ -0,0 +1,160 @@
...
+end
+
+function anomaly_score_gpu(ae::AE, real_input; batch_size=64, dims=(1,2,3), to_testmode::Bool=true)
+	real_input = Flux.Data.DataLoader(real_input, batchsize=batch_size)
+	(to_testmode == true) ? Flux.testmode!(ae) : nothing
+	ae = ae |> gpu
+	scores = Array{Float32}([])
+	for x in real_input
+		x = x |> gpu
+		score = vec(Flux.sum((ae(x) .- x).^2, dims=dims))
```
vitskvara: Reuse the previous function here instead doing it all over again. When changing stuff later, you only have to make the change in one place.

vitskvara: Also, you are 100% sure that the code is consistent.

src/models/ae.jl
```diff
@@ -0,0 +1,160 @@
...
+		scores = cat(scores, score |> cpu, dims=1)
+	end
+	(to_testmode == true) ? Flux.testmode!(ae, false) : nothing
+	return scores
+end
+
+
+function StatsBase.fit!(ae::AE, data, params)
+	# prepare batches & loaders
+	train_loader, val_loader = prepare_dataloaders(data, params)
```
vitskvara, line 90: What is this?

vitskvara (changes requested):

src/models/ae.jl
TheMrGhostman: ok, i will fix that

scripts/experiments_images/DeepSVDD.jl
TheMrGhostman: SVDD is, in first stage of training, only AE with additional parameters. So it was mainly naming issue at that time

src/models/ae.jl
TheMrGhostman: I don't understand. Consistent in what way? You mean anomaly_score and anomaly_score_gpu?

src/models/ae.jl
TheMrGhostman: you mean `train_loader, val_loader = prepare_dataloaders(data, params)`?

It is just function that creates dataloaders for training 

```julia
function prepare_dataloaders(data; batch_size::Int=64, iters::Int=10000)
	train_loader = MLDataPattern.RandomBatches(data[1][1], size=batch_size, count=iters)
	# for cheching convergence I need to drop anomal samples from validation data
	val_data_ind = findall(x->x==0, data[2][2])
	if length(size(data[2][1]))==4
		val_data = data[2][1][:,:,:,val_data_ind]
	else
		val_data = data[2][1][:,val_data_ind]
	end
	val_loader = Flux.Data.DataLoader(val_data, batchsize=batch_size)
	return train_loader, val_loader
end
```

src/models/ae.jl
vitskvara: I would define a separate one line function
```
mse_score(m,x,dims) = vec(Flux.sum((m(x) .- x).^2, dims=dims))
```
and reuse it in the two functions.

